PR #5, opened 2014-01-16
Added a README and updated tests slightly to better reflect more dynamic...

ghost:
... cases. Also added an option to enable the fancy print output(disabled by default so ginkgo can run without erasing the messages).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Run
 To run this project, you will first need to go the "Setting up PATs" section. Afterwards, you can
 change into the pat directory and run:
 
-1) Go therough the "Setting up Go" section
+1) Go through the "Setting up Go" section
 
 2) change into the top level of this project
 
@@ -63,7 +63,7 @@ change into the pat directory and run:
 
 	# go run pat/main.go
 
-OR install the binary
+OR install the binary (recommended)
 
 2) Change into the main directory
 
@@ -73,5 +73,20 @@ OR install the binary
 
 3) Run PATs from the command line
 
-	# pat 
+	# pat
+
+4) Run PATs as an HTTP server (work in progress)
+
+	# pat -server
+
+Example calls:
+
+	# pat -h [will output all of the command line options if installed the recommended way]
+
+	# pat -concurrency=5 -pushes=5 [This will start 5 concurrent threads all pushing 1 application]
+
+	# pat -concurrency=5 -pushes=1 [This will only spawn a single concurrent thread instead of the 5 you requested because you are only pushing a single application]
+
+	# pat -silent [if you don't want all the fancy output use the silent flag]
+ 
 
--- a/README.md
+++ b/README.md
@@ -1,0 +1,77 @@
+PATs (Performance Acceptance Tests)
+==================================
+The goal of this project is to create a load generation testing framework.
+
+
+Setting up PATs
+==================================
+To setup this project, a number of requires need to be met for GO.
+
+1) Ensure that GO (64bit version) has been installed on the system.
+
+2) Setup the GOPATH
+
+        # export GOPATH=~/go (or any other workspace repository for all your go code)
+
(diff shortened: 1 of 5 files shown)


system_test.go
```diff
@@ -27,7 +32,7 @@ var _ = Describe("System", func() {
 
 		It("Reports app push speed correctly", func() {
 			json := post("/experiments/push")
-			Ω(json["TotalTime"]).Should(BeNumerically("~", 1*time.Second, 2*time.Second))
+			Ω(json["TotalTime"]).Should(BeAssignableToTypeOf(json_time))
```
julz, line 35: this isn't really testing much now.. admittedly BeNumerically wasn't really working either :)... I guess we can leave this for now and when we have support for different scenarios the tests will start being much easier to write

pat/main.go
```diff
@@ -10,6 +10,7 @@ func main() {
 	server := flag.Bool("server", false, "true to run the HTTP server interface")
 	pushes := flag.Int("pushes", 1, "number of pushes to attempt")
 	concurrency := flag.Int("concurrency", 1, "max number of pushes to attempt in parallel")
+	output := flag.Bool("output", false, "true to run the commands and print output the terminal")
```
julz: for the flag, maybe "-silent" to not have the output (defaulted to false)? I think showing output should be the default (because it's what most people will want if they're calling via the cli)

README.md
```diff
@@ -0,0 +1,77 @@
...
+4) Run all tests within the repository
+
+	# not sure how to do this yet but we should
+
+Run
+==================================
+To run this project, you will first need to go the "Setting up PATs" section. Afterwards, you can
+change into the pat directory and run:
+
+1) Go therough the "Setting up Go" section
```
julz: therough->through

README.md
```diff
@@ -0,0 +1,77 @@
...
+
+	# go run pat/main.go
+
+OR install the binary
+
+2) Change into the main directory
+
+	# cd $GOPATH/src/github.com/julz/pat/pat
+
+	# go install
```
julz, line 72: I _think_ this requires you to have set PATH to $PATH:$GOPATH/bin first?

README.md
```diff
@@ -0,0 +1,77 @@
...
+
+3) Run PATs from the command line
+
+	# pat
```
julz: -> 

pat -concurrency=[CONCURRENCY] -pushes=[PUSHES]
pat -help

4) Run PATs as an HTTP server (work in progress)

pat -server
open http://localhost:8080/experiments/

system_test.go
```diff
@@ -9,17 +9,22 @@ import (
...
 	Describe("Running PATs with a cmd line interface", func() {
-		It("Runs a push and responds with the speed", func() {
-			output := RunCommandLine(1, 1)
-			Ω(output.TotalTime).Should(BeNumerically("~", 1*time.Second, 2*time.Second))
+		It("Runs a single push and responds with the speed", func() {
+			output := RunCommandLine(1, 1, false)
+			Ω(output.TotalTime).Should(BeAssignableToTypeOf(push_time))
 		})
+
+		PIt("Should fail to push an app when not logged in and report an error", func() {})
```
julz, line 24: hmm good point, how do we do auth? that's probably a Story so we can think about the best way to do it

README.md
ghost, line 72: 1) "Setting up PATs" should take care of this. Guess I can add a note that GOPATH and PATH need to be setup again if they close the terminal and reopen it.

julz, line 72: Oh, no, you're right, my bad

system_test.go
ghost, line 35: I would like to do something like Ω(err).Should(BeNil()) to go along with this test, but the errors are not being bubbled up to the top right now, or we need to have some way to grab them. Figured that would be a future concern.

PR #6, opened 2014-01-16
Initial support for user clustered workers

julz:
Adds a Worker interface with two implementations, LocalWorker runs the test in a Goroutine, redis.Worker send a redis message to trigger the benchmark on a remote node. redis.Slave has some code to pull tasks off the queue, benchmark them, and then reply via redis with the result of the benchmark. 

Can probably do this much better, e.g. with one reply connection feeding in results rather than a reply connection for each slave, but this is a start.


julz:
@DanLavine think your comments are on the first commit and were already changed/fixed by the second? Still, this shows there needs to be an integration test for these two methods, I'll add one

PR #10, opened 2014-01-18
Add CSV download link to UI, backfill more tests, add d3 table in UI

julz:
Finishes 2098

### DIFF
--- a/cmdline.go
+++ b/cmdline.go
@@ -1,11 +1,9 @@
 package pat
 
 import (
-	"encoding/csv"
 	"fmt"
 	"github.com/julz/pat/experiment"
-	"os"
-	"strconv"
+	"github.com/julz/pat/output"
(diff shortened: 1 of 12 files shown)


cmdline.go
```diff
@@ -1,11 +1,9 @@
 package pat
 
 import (
-	"encoding/csv"
 	"fmt"
 	"github.com/julz/pat/experiment"
-	"os"
-	"strconv"
+	"github.com/julz/pat/output"
```
simonleung8, line 6: Can't find package output, seems like it doesn't exist?

julz, line 6: Damn, must not have "git add"ed the folder, not near a computer, will check in a couple hours and update - sorry!

julz, line 6: ok, done, should work now I hope!

PR #16, opened 2014-01-21
Allows an experiment to be continuously repeating at n seconds interval

simonleung8:
Finished 2085
- Call Repeat(second, func) with second > 0 will spawn a goroutine to run task continuously with n second interval. It returns a *RepeatItem, which allows stopping the repeating timer e.g. RepeatItem.Stop()
- Repeat(0, func) will run the task immediately without using gorountine, and nil will be returned

### DIFF
--- a/interval/interval_test.go
+++ b/interval/interval_test.go
@@ -19,7 +19,7 @@ var _ = Describe("Interval", func() {
 		}
 	})
 
-	Describe("Repeat returns nil or *ReturnItem depends on the value of second", func() {
+	Describe("Repeat()", func() {
 
 		Context("When Repeat() is called with seconds and fn() provide", func() {
 
--- a/interval/interval_test.go
+++ b/interval/interval_test.go
@@ -1,0 +1,11 @@
+package interval_test
+
+import (
+	. "github.com/julz/pat/interval"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Interval", func() {
+
+})
--- a/interval/interval_test.go
+++ b/interval/interval_test.go
@@ -7,5 +7,23 @@ import (
 )
 
 var _ = Describe("Interval", func() {
+	//backfill more tests
+	
+	Describe("Repeat returns nil or *ReturnItem depends on the value of second", func() {
+		testFunc := func() {
+			return
+		}
+		
+		Context("returns nil when seconds <= 0", func() {
+			var second = 0
+				Ω(Repeat(second, testFunc)).Should(BeNil())
+		})
 
+		Context("returns *RepeatItem when seconds > 0", func() {
+			var ptr *RepeatItem
+			var second = 1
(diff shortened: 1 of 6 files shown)


simonleung8:
blah. Forgot 'go fmt'

julz:
Simon - good tip is to use "git add -p" instead of "git add" then you can see if your changes seem like what you expect. Helps to catch forgetting to fmt etc.

Anyway can you be a bit more careful with your tests and preferably write the code test-first? None of those tests are getting run because they're in context(..) blocks not it(..) blocks. If you write test first you'll know your tests work because you'll see them fail before you write the code.

Also a) don't like returning nil, it's probably going to cause a bug, b) try to make the tests describe how I use the code, not just repeat what the code says. We can chat later about writing tests / pair for a bit if you like?

simonleung8:
Ah, did not notice the tests were in the wrong blocks, but all the tests were actually being run because they were failing with a "nil pointer dereference" error when I put the var testFunc in 'BeforeEach' block.  I thought it was weird but the tests being in the wrong block must have been the reason why.
Anyway, let me correct it and put in more tests.

interval/interval_test.go
```diff
@@ -0,0 +1,35 @@
...
+	})
+
+	Describe("Repeat()", func() {
+
+		Context("When Repeat() is called with seconds and fn() provide", func() {
+
+			It("Should return a pointer to RepeatItem", func() {
+				var ptr *RepeatItem
+				var second = 1
+				Ω(Repeat(second, testFunc)).Should(BeAssignableToTypeOf(ptr))
```
julz, line 29: this isn't testing anything except that the function returns basically anything, and I don't think this functionality warrants a whole package anyway. 

We already have `Repeat(x, fn)` in benchmarker.go, and we have `ExecuteConcurrently` and `Execute` to execute arbitrary channels of functions. How about we just create a `RepeatEveryUntil(seconds, stopTime, fn)` method in benchmarker.go similar to the existing `Repeat(x, fn)` and use that? 

In the test for that, let's test that it sends the function 'fn' to the returned channel the number of times you'd expect given the stopTime and the amount of seconds between repetitions, and that the times between repetitions are (approximately) what we expect.

pat/main.go
```diff
@@ -35,6 +36,8 @@ func main() {
 		server.Serve()
 		server.Bind()
 	} else {
-		pat.RunCommandLine(*pushes, *concurrency, *silent, *output)
+		_ = interval.Repeat(0, func() {
```
julz, line 39: your one test for `interval.Repeat(..)` is the value it returns. And then you ignore that value in the one place you call the method :-).

julz:
Pretty sure the formatting still isn't fixed which makes it hard to see the diffs. Can you diff locally rather than pushing lots of commits to github, or even just rebase and squash when you're done, don't want 50 formatting commits in the log :) Ping me if you need help squashing the commits or anything

julz:
this is superseded by the other branch now, right? should we close this pr?

simonleung8:
this is superseded by Pull #22

PR #20, opened 2014-01-22
Add a list of previous runs to the UI, no persistence yet

julz:
Also doesn't hide the start button when a run is triggered, and adds hash-based real URLs for viewing experiments that can be bookmarked


simonleung8:
neat stuff =)

PR #21, opened 2014-01-23
Operations

ghost:
Updates to reflect the command -experiments="..." that are separated by a command line. These changes can also be seen in the cli UI as it monitors what the commands you told the program to execute. 

Also I was not sure how to handle the case where we do -experiments=... -pushes=... -concurency=... So if experiments is provided, there will only be a concurrency of 1 atm.


benchmarker/benchmarker.go
```diff
@@ -41,15 +49,15 @@ func Counted(out chan<- int, fn func()) func() {
 	}
 }
 
-func Timed(out chan<- time.Duration, errOut chan<- error, experiment func() error) func() {
-	return TimedWithWorker(out, errOut, NewWorker().withExperiment("*", experiment), "*")
+func Timed(out chan<- BenchmarkResult, errOut chan<- error, experiment func() error, gcfCommand string) func() {
```
julz: definitely let's not expose the idea that this last parameter is a 'gcfParameter' here. Benchmarker doesn't know about gcf (and we're just about to start a story about being able to use the rest api instead). I think we should keep Timed as-is, if you want more control over how we time something, use TimedWithWorker and pass a worker that acts however you like (see my comments about using a CompositeWorker when we want to execute a series of tasks elsewhere). What do you think?

benchmarker/benchmarker.go
```diff
@@ -5,8 +5,13 @@ import (
 	"time"
 )
 
+type BenchmarkResult struct {
+	Command string
+	Time    time.Duration
```
julz: we should name the field 'duration' not time. Because it's a time.Duration, not a time.Time.

cmdline.go
```diff
@@ -26,6 +26,22 @@ func RunCommandLine(pushes int, concurrency int, silent bool, name string) error
 	return experiment.Run(pushes, concurrency, output.Multiplexer(handlers).Multiplex)
 }
 
+func RunCommandLineExperiments(silent bool, name string, experiments string) (err error) {
```
julz: why a separate function rather than extending the existing one above? I don't like the duplication we end with between these two methods, at the very least the common code needs to be shared. I think it's better just to use the existing `RunCommandLine` with maybe just an extra if/else block at the bottom.

config/pat_config.go
```diff
@@ -0,0 +1,62 @@
...
+		Server      bool
+		Pushes      int
+		Concurrency int
+		Silent      bool
+		Output      string
+		Experiments string
+	}
+
+	Workloads struct {
+		//dan I think we could do workloads in the futures so structire it like this for now. can remove before merge if needed
```
julz: yeah let's lose it and add it when we need it

config/pat_config.go
```diff
@@ -0,0 +1,62 @@
...
+	"flag"
+	"io/ioutil"
+	"launchpad.net/goyaml"
+)
+
+type Config struct {
+	CliParams struct {
+		Config      string
+		Server      bool
+		Pushes      int
```
julz: Think we should call this 'Iterations' now

experiment/experiment.go
```diff
@@ -29,29 +30,47 @@ type Sample struct {
...
-func Track(samples chan *Sample, results chan time.Duration, errors chan error, workers chan int) {
+func RunExperiments(operations []string, tracker func(chan *Sample)) error {
+	result := make(chan BenchmarkResult)
+	errors := make(chan error)
+	workers := make(chan int)
+	samples := make(chan *Sample)
+	go Track(samples, result, errors, workers)
+	go tracker(samples)
+
+	for _, operation := range operations {
```
julz: commented on the commit, but, in short, do this inside the Execute(..), not outside of it, probably just by having a more intelligent Worker that understands csvs, there should be one Execute per Tracker (maybe we should add some code to check that and throw an error)

experiment/experiment.go
```diff
@@ -29,29 +30,47 @@ type Sample struct {
...
+		Execute(Once(Counted(workers, Timed(result, errors, experiments.Dummy, operation))))
+	}
+
+	time.Sleep(1 * time.Second) //dan The execute function returns to quickly for the UI to update with the last command. ExecuteConcurrently(1 ...) works because the simple checks in wg gives a little more time for the UI to update. Needs some thought.
```
julz: this won't be needed if you make sure there's only one Execute(..) call per tracker

benchmarker/benchmarker.go
ghost: The only reason I had changed it originally is right now we are using it to issue the "push" or "dummy" command, but that is represented by a "*". I can write the original caller to use the worker directly.

experiment/experiment.go
ghost: This is still a problem. Take the original func Run(..) and change the ExecuteConcurrently(...) to Execute(...). The last command probably wont be displayed(could get a lucky timer that it does)

julz: ohh I think I see what you mean, yes, the program will exit before then, which is a sort-of/kind-of race condition because we actually wanted that print message to happen before the program quit. It's fine (I think) for the server version, though. 

I pushed a branch (add_quit_channel) a little bit ago adding a quit channel to experiments so the tracker will stop cleanly when RunExperiments completes. I think we just need to make sure RunExperiments doesn't return until tracker() returns, either with a wait group, or with a done channel, e.g. we could (probably) change L56 to:

```
trackerdone := make(chan bool)
go func() { tracker(samples); trackerdone <- true }
```

and then, at the bottom of RunExperiment do `<-trackerdone`. That way the function doesn't return until all the various handlers (e.g. printing the output/writing the CSV) have completed. What do you think?

ghost:
Alright with these updates I fixed the comments and I had started doing some refactor work but, tossed that in the end. Way to much for this PR. The note about composite workers is a good idea, but we still need a way to pass multiple functions to the composite worker so i added "worker.AddExperiment" function that used the current worker in the end. This allows a single worker to issue multiple commands which is what I think we would want. I would like to refactor stuff so the command "pushes" = "iterations." This way we can say want to concurrently run N iterations and here is a single iteration workload. Also I was trying to come up with a nice error object that could be passed around rather than standard error. This would allow for nicer testing. Not sure what you think about all that. Unfortunately Simon and I went a little slow this week, getting a better understanding of how all these multiple go routines are interacting with multiple channels and everything. Not the easiest thing to do when learning a new language. But if everything is looking better, I would like to try and quickly get another work item done this friday/weekend.

julz:
Hey Dan - that all sounds really good. I was also thinking of renaming pushes->iterations and repeating the set of experiments as a set. That should scale nicely to more complicated scripts if we later want those. 

Re: the error object, I think given this is Golang just think in interfaces.. We don't need to construct a particular object/struct type, we just need to have an interface with any methods, and then pass anything that conforms to that. Basically, no need to move away from using standard error until there's a method we need that standard error doesn't have. (Is there?)

Don't worry too much about the slowness, you guys are doing really well coming up to speed, and it's normal for our velocity to start slow and improve as we get the hang of stuff and as the architecture solidifies and we understand how it all fits together. I also think as that happens we'll start to refactor stuff down a bit, for example there really are way too many channels going on, and I think next time I touch that code I'll try to make it a bit clearer what's going on.

julz:
ANYWAY :-), sounds like you don't want this branch merged yet, so let me know when I should look at it again

benchmarker/benchmarker.go
```diff
@@ -17,20 +22,28 @@ func NewWorker() *LocalWorker {
 	return &LocalWorker{make(map[string]func() error)}
 }
 
+func (self *LocalWorker) AddExperiment(name string, fn func() error) *LocalWorker {
```
julz: this is identical to withExperiment, below. I use the 'with' prefix to indicate that it's chainable, but AddExperiment is just as good. Either way, pick one :-)

cmdline.go
```diff
@@ -12,18 +12,29 @@ type Response struct {
...
 	}
 
 	if len(name) > 0 {
 		handlers = append(handlers, output.NewCsvWriter(name).Write)
 	}
 
-	return experiment.Run(pushes, concurrency, output.Multiplexer(handlers).Multiplex)
+	if experiments != "" {
+		operations := strings.Split(experiments, ",")
+		return experiment.RunExperiments(operations, output.Multiplexer(handlers).Multiplex)
```
julz: I think we can lose this if/else if we just default 'experiments' to "push" (that way it's never empty, and the default case - repeated pushes - is just a very simple single-experiment list)

benchmarker/benchmarker.go
ghost: I was just thinking maybe there could be a reason we would want to keep one of them as a private function vs public function when i was writing it. I'll just leave 1 in and make it public.

benchmarker/benchmarker.go
```diff
@@ -41,18 +53,21 @@ func Counted(out chan<- int, fn func()) func() {
 	}
 }
 
-func Timed(out chan<- time.Duration, errOut chan<- error, experiment func() error) func() {
-	return TimedWithWorker(out, errOut, NewWorker().withExperiment("*", experiment), "*")
-}
-
-func TimedWithWorker(out chan<- time.Duration, errOut chan<- error, worker Worker, experiment string) func() {
+func TimeWorker(out chan<- IterationResult, bench chan<- BenchmarkResult, errOut chan<- error, worker Worker, operations []string) func() {
```
julz: Why did we rename this? The function doesn't time a worker, it uses a worker to time another function.

experiment/experiment.go
```diff
@@ -29,35 +38,54 @@ type Sample struct {
...
 	workers := make(chan int)
 	samples := make(chan *Sample)
 	quit := make(chan bool)
-	go Track(samples, result, errors, workers, quit)
+	go Track(iteration, samples, benchmark, errors, workers, quit)
 	go tracker(samples)
+
+	worker := NewWorker()
+
+	if len(operations) == 1 && operations[0] == "" { //(dan) right now default to push if no operations are provided
```
julz: do we need the comment? it doesn't seem to add anything that's not very clear in the code

ghost:
Squashed and good to go.

PR #22, opened 2014-01-23
Repeat a function at n interval and stop at s seconds

simonleung8:
Repeat function is now in benchmark package, with CLI support

example usage:

//repeat workload every 5 seconds and stop after 60 seconds
go run pat/main.go -interval 5 -stop 60


experiment/experiment.go
```diff
@@ -67,8 +72,14 @@ func (ex *RunningExperiment) run() {
 		select {
 		case result := <-ex.results:
 			sampleType = ResultSample
-			n = n + 1
-			totalTime = totalTime + result
+			if n < int64(total) {
```
julz: not sure what's going on here, can we have a test explaining it?

experiment/experiment.go
```diff
@@ -40,18 +41,22 @@ func Run(pushes int, concurrency int, tracker func(chan *Sample)) error {
 	errors := make(chan error)
 	workers := make(chan int)
 	samples := make(chan *Sample)
-	go Track(samples, result, errors, workers)
+	quit := make(chan bool)
+	go Track(samples, result, errors, workers, quit, pushes)
 	go tracker(samples)
-	ExecuteConcurrently(concurrency, Repeat(pushes, Counted(workers, Timed(result, errors, experiments.Dummy))))
+	//ExecuteConcurrently(concurrency, Repeat(pushes, Counted(workers, Timed(result, errors, experiments.Dummy))))
+	Execute(RepeatEveryUntil(6, 20, func() {
```
julz: obviously 6 and 20 need to be externalised before we merge, also lose the commented dead code (that's what git history is for :-). Also need to make sure this code works in the case where the user didn't ask to repeat the experiment on a schedule.. e.g. RepeatEveryUntil(0,0) needs to do something sensible.

simonleung8:
Some new tests, took out dead code as well

simonleung8:
Interval repeat is now run as 1 long experiment. 
new flag in CLI to trigger feature (-interval, -stop)
This feature is not yet exposed in the Web UI

cmdline.go
```diff
@@ -12,25 +12,32 @@ type Response struct {
...
 	if len(name) > 0 {
 		handlers = append(handlers, output.NewCsvWriter(name).Write)
 	}
 
-	return experiment.Run(pushes, concurrency, output.Multiplexer(handlers).Multiplex)
+	return experiment.Run(pushes, concurrency, interval, stop, output.Multiplexer(handlers).Multiplex)
 }
 
-func display(target int, concurrency int, samples chan *experiment.Sample) {
+func display(target int64, concurrency int, interval int, stop int, samples chan *experiment.Sample) {
```
julz, line 29: why did int change to int64?

julz:
looks good, will test out properly and merge later, I might also squash some of these commits together before I push this to master if that's ok?

simonleung8:
Go ahead and squash some of the commit if you want.

julz:
couldn't squash because you'd already merged against master a few times. If possible lets try to avoid merge commits in the logs by rebasing rather than merging? Anyway, merged.

PR #24, opened 2014-01-25
highlight active experiment, and fix calling refresh() too often

julz:
This might well account for the performance problems you were seeing, Simon. We were accidentally calling refresh an increasing numbers of times because the timer wasn't being cancelled. Also formatted the javascript in the same commit, sorry about that.


julz:
superseded by https://github.com/julz/pat/pull/25

PR #25, opened 2014-01-25
highlight active experiment, fix calling refresh() too often AND a start on end-to-end tests

julz:


simonleung8:
I am guessing your IDE's tab or spacing is different than the original files, git seems to think there are changes to the lines while there aren't, it makes the changes harder to see.  Is there a way for us to standardize the formatting.

Other than that, the changes seems good, I will test and merge later.

PR #26, opened 2014-01-26
Dynamic bar chart (no scrolling yet)

simonleung8:
With this small change, the bar chart will update the x and y axises dynamically.  Scrolling is not yet implemented, so depends on screen resolutions, the bars start to overlap each other after certain number of pushes (30+ on mine).

### DIFF
--- a/ui/js/app.js
+++ b/ui/js/app.js
@@ -88,7 +88,8 @@ pat.experimentList = function() {
 ko.bindingHandlers.chart = {
   c: {},
   init: function(element, valueAccessor) {
-    ko.bindingHandlers.chart.b = d3.custom.barchart(d3.select(element));
+    //ko.bindingHandlers.chart.b = d3.custom.barchart(d3.select(element));
+    ko.bindingHandlers.chart.b = d3.custom.barchart(element.id);
   },
   update: function(element, valueAccessor) {
     var data = ko.unwrap(valueAccessor())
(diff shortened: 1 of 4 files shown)


julz:
can we get tests for this please? you can update the graph data (in the test) and check the axis have the expected labels and the bars are correctly spaced based on the data you pass (just check the DOM for the svg objects you expect to exist, e.g. `expect($("svg rect.bar").length).toBe(5)` etc.

simonleung8:
Added some tests for bar chart.

Bar chart will show at least 10 pushes(tasks) by default, and it will start scaling if there are more than 10 pushes

ui/js/app.js
```diff
@@ -88,7 +88,7 @@ pat.experimentList = function() {
 ko.bindingHandlers.chart = {
   c: {},
   init: function(element, valueAccessor) {
-    ko.bindingHandlers.chart.b = d3.custom.barchart(d3.select(element));
+    ko.bindingHandlers.chart.b = d3.custom.barchart(element.id);
```
julz: hey, this all looks good but I'm not sure about this line.. it'll break if the graph container doesn't contain an id (and there's no reason the graph container needs to have an id). How about passing just `element` (i.e. `d3.custom.barchart(element)`) and doing `d3.select(el)` in line 2 of `d3.custom.barchart`? d3.select on a node returns the node, and this would also support passing an id (as `"#foo"`), because d3.select on a string does a query.

julz:
merged manually

PR #28, opened 2014-01-30
Hide x-axis ticks in bar chart when # of workload is <10

simonleung8:
Currently the bar chart shows ticks 1 to 10 by default even if the # of workload is <10.

With this change, the default size of x-axis remain the same (10 workloads), but the chart will not show extra ticks in the x-axis when workload is <10.

### DIFF
--- a/ui/js/app.js
+++ b/ui/js/app.js
@@ -43,7 +43,7 @@ pat.experiment = function(refreshRate) {
       exports.csvUrl(data.CsvLocation)
       exports.refreshNow()
       })
-    }	
+    }
   }
 
   exports.view = function(url) {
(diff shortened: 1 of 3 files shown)


julz:
I don't think this is the right approach. If there're more than 10 results we should have less frequent ticks, but we shouldn't turn them off entirely. Also, one commit for both code and test please (ideally, the test should be written first, but definitely they should be in the same commit).

simonleung8:
This change does not hide the ticks when there are more than 10 results, it only hides the extra ticks when the results are less than 10.

Also, squashed into 1 commit

simonleung8:
Superseded by #34

PR #30, opened 2014-01-31
Extract logic from server in to Laboratory class

julz:
Now single laboratory object (a Laboratory is somewhere experiments are
run) is used by both cmd line and server and has all the logic that had
crept in to server.go.


julz:
@DanLavine I agree, especially I don't like that it's duplicated right now, but we don't support anything more than these functions right now so no need to make it much more complicated just yet.. once we get the story about being able to use REST api directly rather than just via gcf (or a story about supporting pushing specific apps) I think it'll make sense to do this better

julz:
@DanLavine commit above fixes the bug you were seeing, command line was exiting too quickly.

ghost:
Alright, I have gone through the code and some of it seems a bit odd to me. For example, there is the "type Experiment interface" but those functions are not defined in the Experiment.go class. They are defined in the Laboratory.go class. Also we cannot use the -output flag anymore everything is given a random string csv file name. With such a large change, think we should have walked through some of the code and logic over pair programming sessions.

julz:
@DanLavine no test broke when I removed the output setting, so I assumed I hadn't broken anything. If that's a requirement it's easy to add back in, but let's put tests around things so we can refactor safely. The main point of this refactoring was to reduce code duplication and (massively) increase test coverage. I still have a little more to go, e.g. to get coverage around cmdline.go and some of the inner stuff in runner.go/Execute, but moving the logic out to laboratory is what makes that possible. The Experiment interface is defined in the experiment package, exactly as you'd expect (in a file runner.go, but that's not really important, happy to rename it back, originally I renamed it because I was going to have laboratory also in the experiment package, but I changed my mind about that). 

Re: pairing, I'm a bit fan of pairing but obviously given time zones we can't do it for every change, and it's really important to be able to refactor and improve things all the time. That's why I insisted we do code review via pull requests rather than committing directly when we're not pairing. So: what things would you like to change or improve?

PR #31, opened 2014-02-04
Simple tooling that allows a user to view the current push per second averages over a single run.

ghost:
Currently the pushes per second (or commands per second as it will time the dummy function "login" as well) will provide useful information for a static single dry run. Currently if you run PATs with the interval setting the data will become skewed. The reason for this is PATs does not have a wall time for individual commands. Also, running a workload say for example: login,login,push,login will skew the data. We need a more concrete way to define how long it took to run the login commands vs the push command. Reason for this is that push will have a much slower response time than login, but a login at the end of the command queue will report the 3 login times as login(3) + push. However, currently our UI only allows for the ability to issue one command type at a time and no repetition. So this functionality could be implemented to complete the UI story portion.

### DIFF
--- a/cmdline.go
+++ b/cmdline.go
@@ -67,11 +67,12 @@ func display(concurrency int, iterations int, interval int, stop int, samples <-
 		fmt.Println()
 		for key, command := range s.Commands {
 			fmt.Printf("\x1b[1m%v\x1b[0m:\n", key)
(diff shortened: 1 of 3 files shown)


julz:
hey, looks good, didn't like the tests waiting on timers so I injected the ticker instead which means the tests can dummy out the actual time and not have to rely on physics

PR #33, opened 2014-02-05
updates to throughput calculations and data.

ghost:
Updates to how we are saving data and recording throughput. Much more dynamic and allows for custom query sets on any number of commands for specific time intervals.

### DIFF
--- a/experiment/runner.go
+++ b/experiment/runner.go
@@ -1,9 +1,10 @@
 package experiment
 
 import (
+	. "github.com/julz/pat/benchmarker"
+	"math"
 	"strings"
 	"time"
-	. "github.com/julz/pat/benchmarker"
 )
 
 type SampleType int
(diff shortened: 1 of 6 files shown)


experiment/runner.go
```diff
@@ -143,12 +150,14 @@ func (ex *ExecutableExperiment) Execute() {
 	Execute(RepeatEveryUntil(ex.Interval, ex.Stop, func() {
 		ExecuteConcurrently(ex.Concurrency, Repeat(ex.Iterations, Counted(ex.workers, TimeWorker(ex.iteration, ex.benchmark, ex.errors, ex.Worker, operations))))
 	}, ex.quit))
-
+	time.Sleep(1 * time.Second) //(dan) need to halt for the last throughput to catch data
```
julz: I don't think we need this line any more (the command line waits for the user to quit), and if we do we need to not.

julz:
I took ages to look at this, sorry; can you rebase it on top of master when you get a chance so travis runs against it?

experiment/runner.go
ghost: The process listening for all the incoming messages is killed by the next line "close(...)." This means that the last few results are never accounted for when determining the throughput because that runs on an iterative loop each second. To account for that, I will have to wait a second for the throughput information to update.

julz: mmm.. close just closes the results channel, the Sample() method will still receive all the remaining results on the channel before it finally receives the close and completes processing. If we need to ensure that we also process another channel before returning from the function, then we need to do that, not shove the mess under the bed with a sleep statement :)

julz: in https://github.com/julz/pat/pull/37 I moved BechmarkResult (renamed to StepResult) and the Errors channel in to IterationResult so we just have one stream of results to process. Think that simplifies things a lot (at the small price of not receiving the steps until the iteration completes, which is what JMeter does also). I think that might also remove the problems with this method finishing early, because we only have one channel to wait for, and we already process all the Iteration results before returning.

julz:
this has been open for ages and I think it's now being done in another branch, closing

PR #34, opened 2014-02-06
New bar chart - more dynamic in scale and allows panning left and right

simonleung8:
This new bar chart has panning feature, the code also set the stage for zooming (if needed in the future)


julz:
stupid question... panning looks nice, but why are the bars going from right to left now?

simonleung8:
I originally have the bars left to right and the chart will auto pan when bars go out of screen.  I rather have the user control the pan themselves and the chart stays at pan(0), the user will able to see the latest data without panning if the bars go right to left.  Also if user flipping through histories, the latest data will always be there at pan(0).

julz:
can we have this functionality but going in the other direction? I think people naturally assume that the time axis on a graph progresses from left to right..

simonleung8:
bars are now left to right.  
I know there are rooms to improve the chart, let's wait and see what we need after all other charts are finished.

ui/index.html
```diff
@@ -134,7 +135,9 @@
 
   <!-- **************** footer ******************* -->
   <script>
-ko.applyBindings(new pat.view( new pat.experimentList(), pat.experiment(800) ));
+    var patview = new pat.view( new pat.experimentList(), pat.experiment(800) );
+    var barchart = new d3.custom.barchart(document.getElementById("graph"), patview.data);
```
julz: what's the thinking behind pulling this out here? I quite like that the view was being done declaratively via binding but now it's hardcoded in javascript that this particular id is linked to this particular part of the view. What was wrong with the previous approach?

ui/js/bar.js
```diff
@@ -1,57 +1,185 @@
...
+  this.y = d3.scale.linear().range([0, this.svgHeight], 1);
+  this.xAxis = d3.svg.axis()
+    .scale(this.x)
+    .orient("bottom")
+    .ticks(0);
+  this.yAxis = d3.svg.axis()
+    .scale(this.y)
+    .orient("right")
+    .tickSize(-this.svgWidth + 30);
+  this.zoom = d3.behavior.zoom()
```
julz, line 25: is this being used?

ui/js/bar.js
```diff
@@ -1,57 +1,185 @@
...
+          self.data = observableArray();
+          self.refresh()();
+      } else {
+        self.data = observableArray();
+        self.refresh()();
+      }
+  });
+};
+
+d3.custom.barchart.prototype.refresh = function() {
```
julz: I don't think you need this. This function is already bound to the changing data, so you don't need to pass in a subscribable array so that you can resubscribe again. Im guessing this is why you had to break the data binding in index.html.

ui/js/tests.js
```diff
@@ -105,57 +105,39 @@ describe("Throughput chart", function() {
...
-    }
-    chart(tasks);
-    expect( chart.xAxis_max() ).toBe(10);
-  });
-
-  it("should scale and show >10 tasks in the x-axis if >10 tasks has been performed", function() {
-    var svg = d3.selectAll(".barchart");
-    var tasks = [];
-
+  it("should scale and show >10 tasks in the x-axis if >10 tasks has been performed", function() {
```
julz: this has nothing to do with >10 any more

julz:
this doesn't seem to work right on Safari. The tick labels on the bottom jump around like disney characters and assume weird orders at certain points while they try to animate around. When new rows arrive they whizz out from the right in an odd way, while the matching labels whizz out from the far left. See below, or try it out in safari.

pasted_image_07_02_2014_14_07-2

pasted_image_07_02_2014_13_36-2

simonleung8:
just pushed up the changes to change the animation behavior a bit, the labels should not be dancing around all that much now.

simonleung8:
just pushed up some new changes.
- changed how bar is drawn
- code refactor
- new tests

julz:
still some refactoring to do on this; I don't want to expose observables to the graph (see my comments inline), it's already set up fine as it was -- we don't need two levels of observables (the graph and the binding) and the extra subscribe/refresh function. Also there's still dead code in there - e.g. stuff for zooming on L24. The new test looks good, but I still think there's more code than we actually need for the function (probably a fair bit of that is the extra code for subscribe/refresh that should be in the binding).

simonleung8:
using the old method for data binding now :)

ui/js/bar.js
```diff
@@ -1,57 +1,213 @@
...
+      self.barCon.attr("transform", "translate(" + d3.event.translate[0] + ",0)scale(1, 1)");
+    }
+  }
+
+  
+  var bar_width = function() {
+    return this.barWidth;
+  }
+
+  function reBoxing(obj) {
```
julz: can we change the name of this? I have no idea what reBoxing means

ui/js/bar.js
```diff
@@ -1,57 +1,213 @@
-d3.custom.barchart = function(el) {
-  var d3Obj = d3.select(el);
-  var jqObj = $(el);
-  var xAxis, yAxis
-  d3Obj.append("svg").attr("class","barchart").attr("width", jqObj.width() - 20).attr("height", jqObj.height() - 20);
-
-  function barchart(data) {
-    if (data.length === 0) return;
+var barchart = function(el, observableArray) {
```
julz: you still have observableArray here, even though it's now unused

ui/js/bar.js
```diff
@@ -1,57 +1,213 @@
...
+  function zoomed() {
+    var self = this;
+    return function() {    
+      var svg = self.svg;    
+      self.barCon.attr("transform", "translate(" + d3.event.translate[0] + ",0)scale(1, 1)");
+    }
+  }
+
+  
+  var bar_width = function() {
```
julz: this function is unused now

ui/js/bar.js
```diff
@@ -1,57 +1,213 @@
...
   }
 
-  barchart.xAxis_max = function() { return xAxis.scale().domain()[1]; }
-  barchart.yAxis_max = function() { return yAxis.scale().domain()[0]; }
+  function transformChart(node, x, y) {
+    node.transition()
+      .attr("transform", "translate(" + x + ", " + y + ")" );   
+  }
+
+  function addjustZoomX(zoom, x, range) {
```
julz: addjust -> adjust

ui/js/bar.js
```diff
@@ -1,57 +1,213 @@
...
+      var bars = self.barCon.selectAll("rect.bar").data(data);
+      var labels = self.barCon.selectAll("text").data(data);
+
+      var reboxObj = {
+        self:self,
+        chartBodyG: self.barCon,
+        drawBoxWidth: self.drawBoxWidth,
+        zoom: self.zoom
+      };    
+      reboxObj = reBoxing(reboxObj);
```
julz: what the hecklers are these two lines (100 & 101) doing and why? :)

ui/js/bar.js
```diff
@@ -1,57 +1,213 @@
...
+
+      bars.exit().remove();   
+      labels.exit().remove();
+    } //end returned function
+
+    chart.yAxis_max = function() { return self.yAxis.scale().domain()[0]; };
+    chart.xAxis_max = function() { return self.xAxis.scale().domain()[0]; }
+    chart.draw_box_width = function() { return self.drawBoxWidth; }
+
+    return chart;
```
julz: why are we returning a function from within a function returning a function? Does Alice know about this? The idea of an 'exports' function being returned is that we can use it to 'export' public fields and methods. I don't know why we're then returning another sub method inside that. What's the value of the double nesting? Can we lose it?

ui/js/bar.js
```diff
@@ -1,57 +1,213 @@
...
+      self.outterBody.select(".x.axis").call(self.xAxis);    
+      self.outterBody.select(".y.axis").call(self.yAxis);
+
+      bars.exit().remove();   
+      labels.exit().remove();
+    } //end returned function
+
+    chart.yAxis_max = function() { return self.yAxis.scale().domain()[0]; };
+    chart.xAxis_max = function() { return self.xAxis.scale().domain()[0]; }
+    chart.draw_box_width = function() { return self.drawBoxWidth; }
```
julz: can we pick either camelCase or underscore_words and stick to it?

ui/js/bar.js
```diff
@@ -1,57 +1,213 @@
...
+      .attr("transform", "translate(" + x + ", " + y + ")" );   
+  }
+
+  function addjustZoomX(zoom, x, range) {
+    zoom.x(d3.scale.linear().range([1, range]));                     
+    zoom.translate([x, 0]);
+    return zoom;
+  }
+
+  return exports();
```
julz: just `return exports` would do, and then you can move the code from 'chart' into the top of 'exports', you can then move `var self` on line 85 to point at the outer object, and use 'exports' to point at the publicly exposed fields.

simonleung8:
slimmed down the code a little bit more ~

ui/js/bar.js
```diff
@@ -1,57 +1,191 @@
...
+    .orient("bottom")
+    .ticks(0);
+  this.yAxis = d3.svg.axis()
+    .scale(this.y)
+    .orient("right")
+    .tickSize(-this.svgWidth + 30);  
+  this.zoom = d3.behavior.zoom()
+    .x(this.x)
+    .scaleExtent([1, 10])
+    .on("zoom", zoomed(this));
```
julz: why don't we just make this an inline closure? e.g. 

```
.on("zoom", function() {
  self.barCon.attr("transform", "translate(" + d3.event.translate[0] + ",0)scale(1, 1)");
})
```

Seems clearer than passing `this` down to a one-line function

ui/js/bar.js
```diff
@@ -1,57 +1,191 @@
...
+    .attr("transform", "rotate(-90)")
+    .text("Seconds");
+  this.svg.append("text")
+    .attr("x", this.svgWidth - 60)
+    .attr("y", this.svgHeight + 25)
+    .text("Task #")
+    .attr("text-anchor","middle");
+
+  self = this;
+  var exports = function(data) {
```
julz: indentation a bit wrong below here I think

ui/js/bar.js
```diff
@@ -1,57 +1,191 @@
...
+        .append("text")
+          .attr("x", function(d, i) { return (x(i) + (self.barWidth / 2)) })
+          .attr("y", self.svgHeight + 30 )
+          .attr("dy", ".7em")
+          .text(function(d, i){ return i + 1 })
+        .transition()
+          .attr("y", self.svgHeight + 3 );
+
+      self.outterBody.select(".x.axis").call(self.xAxis);    
+      self.outterBody.select(".y.axis").call(self.yAxis);
```
julz: `outterBody` -> `outerBody`

julz:
better, made some small comments inline

Visually though... do we _Really_ need to animate the axis labels vertically? They seem like they're jumping up from the bottom, which is just.. silly looking. It looks cartoony. Also when the bars appear from the right they seem to come from too far to the right which makes them look more animated than they should (they seem to travel faster when they first appear than the rest of the graph is moving).

Also the graph looks vertically too far down until it has data (at which point it fixes itself), see images below:

pasted_image_12_02_2014_15_42-5

pasted_image_12_02_2014_15_42-2

Apart from that looks much better (and I do like the panning effect apart from the odd animation when new bars are added)

simonleung8:
@julz new commit addressed the things you commented on :)

PR #36, opened 2014-02-07
Added support for deploying PAT to CloudFoundry via a Go Buildpack

jberkhahn:
Added check for VCAP_APP_PORT, default to port 8080 otherwise. Added manifest.yml file for deploying to CloudFoundry in server mode.

Made for use with the go buildpack at github.com/jberkhahn/cloudfoundry-go-buildpack.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,0 +1,3 @@
+applications:
+- name: pat
+  command: pat -server
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"os"
 
 	. "github.com/julz/pat/experiment"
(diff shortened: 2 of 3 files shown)


server/server_test.go
```diff
@@ -135,6 +151,7 @@ func decode(encoded []byte) (decoded map[string]interface{}) {
 }
 
 func req(method string, url string) []byte {
+	fmt.Println("BANANAS: ", url)
```
julz: :) let's lose this line?

manifest.yml
```diff
@@ -0,0 +1,3 @@
+applications:
+- name: pat
+  command: pat -server
```
julz, line 3: I think this should be `go run pat/main.go -server' (so that the UI html files load properly)?

julz:
this is now part of https://github.com/julz/pat/pull/44 right? if so let's close it

PR #37, opened 2014-02-11
Add ability to login and push using REST api

julz:
also simplified some of the benchmarking so we have one channel
with Iterations, Steps and Errors rather than trying to manage
three separate channels. Also moved the handling of multiple steps
in to the worker (this makes more sense because the worker, which might
be on a remote machine, is what is actually responsible for running an experiment).


config/pat_config.go
```diff
@@ -31,6 +33,8 @@ func NewConfig() *Config {
 	flag.StringVar(&config.Workload, "workload", "", "The set of operations a user should issue (ex. login,push,push)")
 	flag.IntVar(&config.Interval, "interval", 0, "repeat a workload at n second interval, to be used with -stop")
 	flag.IntVar(&config.Stop, "stop", 0, "stop a repeating interval after n second, to be used with -interval")
+	flag.StringVar(&config.Username, "username", "", "User to use when operating with REST api")
+	flag.StringVar(&config.Password, "password", "", "Password to use for the REST api user")
```
ghost: Can we add/update a simple tests for these and update the example config-template.yml to give a demonstration on these new features?

experiments/rest.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+type Context struct {
+	loginEndpoint string
+	apiEndpoint   string
+	token         string
+}
+
+func NewContext() *Context {
+	return &Context{"http://login.stage1.ng.w3.bluemix.net/UAALoginServerWAR/", "http://api.stage1.ng.w3.bluemix.net", ""}
```
ghost: Why are these hard coded? We could pass them in through our configuration parser/parameters. Another thing we should consider is passing in a space and organization as well. Will they work if you have more than one of each?

ghost:
Can we pleas add some testing for the rest api. I don't want to leave that as a back story. Also these changes break running pats through the cli.

experiments/rest.go
julz: oh oops, actually these are supposed to be retrieved from /v2/info, my bad, will fix

julz:
hmmm.. I actually was seeing the stuff in experiments/\* as essentially performance test "scripts" which I wasn't going to bother testing, especially since it's an interaction with a remote system which is pretty hard to stub out. I could see an argument for writing some tests around it but I'm not totally convinced they'd be adding value.. let's talk later.

julz:
New commit pushes the config parameters in to the places that actually need them rather than having a super-huge config object passed everywhere. This lets me unhardcode the target/username/password for the rest api, and in the process unhardcode the default CSV directory for the server. Now exposing a DescribeParameters method from experiments which should allow us to consistently expose needed parameters in both cmd line and web ui.

Still todo:
- Expose rest api username/password/target in web ui
- Potentially un-hardcode the namespacing so e.g. "rest:username" would be in that format in the cmdline api, but could be properly indented in the yml / exposed in a section in web ui
- Pull out the common bits of the REST api stuff in to something tested and reusable so we can easily add other methods like delete and scale; potentially might wait until those stories come up to do this though

PR #38, opened 2014-02-12
Implement interval tests from the web UI

jberkhahn:
Implemented  interval tests via webui. Removed blocks in server.go fromaccessing the interval and stop fields from the config generated from the webui. Added ginkgo tests to test these fields. Added fields in
index.html to allow the user to select values for interval and stop, defaulting to 0. Added functions to app.js to package the input into the config that gets sent down. Added tests to test.js to ensure the values are within acceptable bounds.


julz:
looks good to me

PR #39, opened 2014-02-14
previous history popup

simonleung8:
A popup over the chart for all previous histories.


julz:
we probably want to figure out how to set those wait times in the tests lower; no-one likes long running tests, but merging for now

PR #41, opened 2014-02-17
local js libraries

simonleung8:
All JS libraries (JQuery, Knockout, D3) are now included in PAT locally, making this change to avoid having to re-download all the JS files from CDN with every single application run.


julz:
looks good, but can you get rid of ui/js/.DS_Store from the commit? :)

simonleung8:
the ui/js/.DS_Store is actually a "git rm", it is currently in the master and I removed it in this PR

julz:
oh cool, sorry, my bad

PR #42, opened 2014-02-24
Throughput graph for each command

simonleung8:
A graph for showing the throughput per second for each command of an experiment.  

Current limitation:
- This graph cannot show the throughput for experiment histories, as the throughput data is currently not saved.
- As the experiment runs, the graph updates with only the current calculated throughput value, it does not show how the throughput value changes/varies during a running experiment.  It is a limitation of the current data model, which should be updated to be more dynamic soon.
- Currently the graph is a bar chart.  Once the data model is ready, I would like this graph to be a line chart so the throughput changes during a running experiment could be shown.


ui/index.html
```diff
@@ -39,6 +42,17 @@
...
     </div>
     <div id="graph" class="panel-body col-md-12 center-block" data-bind="chart: data">
+      <div class="btn-group-vertical" style="position:absolute; right:-62px; top: 0px;">
+        <button type="button" onclick="dom.contentIn('d3_workload_container');" class="btn btn-default btn-lg">
+          <span class="glyphicon"><img src="images/glyphicons_bars.png" /></span>
+        </button>
+        <button type="button" onclick="dom.contentIn('d3_throughput_container');" class="btn btn-default btn-lg">
+          <span class="glyphicon"><img src="images/glyphicons_lines.png" /></span>
+        </button>
+        <button type="button" onclick="alert('feature coming soon');" class="btn btn-default btn-lg">
```
julz: let's add the button when the feature is actually there, not before

ui/index.html
```diff
@@ -39,6 +42,17 @@
       Experiment Results
     </div>
     <div id="graph" class="panel-body col-md-12 center-block" data-bind="chart: data">
+      <div class="btn-group-vertical" style="position:absolute; right:-62px; top: 0px;">
+        <button type="button" onclick="dom.contentIn('d3_workload_container');" class="btn btn-default btn-lg">
```
julz: no onclick handlers, instead use e.g. `data-bind="click: changeGraph"`, see how the start experiment etc. buttons are done. This will let you test the on-click, and avoid logic in the html.

ui/index.html
```diff
@@ -148,6 +163,8 @@
   <!-- **************** footer ******************* -->
   <script>
     ko.applyBindings(new pat.view( new pat.experimentList(), pat.experiment(800) ));
+    var dom = new DOM();
+    dom.hideContent(["d3_throughput_container"]);
```
julz: this logic should be tested, and should not be in index.html

ui/js/bar.js
```diff
@@ -70,7 +70,7 @@ var barchart = function(el) {
     .attr("transform", "translate(" + (this.svgWidth - 20) + ", 0)");
 
   this.barCon = chartBody.append("g")
-    .attr("id", "chart_box")
+    .attr("id", d3_id + "_box")
```
julz: really not a fan of adding ids like this, it's really brittle and bug-prone (relies on no other component using the same id and only one of a particular component on a page). Better to use classes rather than ids and store and select underneath the passed element.

ui/js/tests.js
```diff
@@ -97,28 +97,65 @@ describe("The view", function() {
   })
 })
 
+describe("DOM elements manipulation", function(){  
+  const bar_chart = "d3_workload_container";
+  const throughput_chart = "d3_throughput_container";
```
julz: way better would be to pass the DOM nodes to swap between rather than ids, then you can just do a generic $(node1) which will work for both nodes and ids.

ui/index.html
```diff
@@ -42,6 +45,14 @@
       </button>
     </div>
     <div id="graph" class="panel-body col-md-12 center-block" data-bind="chart: data">
+      <div class="btn-group-vertical" style="position:absolute; right:-62px; top: 0px;">
+        <button type="button" data-bind="click: function(){changeGraph('workload')}" class="btn btn-default btn-lg">
```
julz: move to view (app.js) and test (don't put logic in the html)

ui/js/throughput.js
```diff
@@ -0,0 +1,149 @@
...
+        .attr("width", jqObj.width())
+        .attr("height", jqObj.height())
+        .attr("class", "throughput")
+      .append("g")
+        .attr("transform", "translate(" + margin.left + "," + margin.top + ")");
+
+    outerBody = svg.append("g");
+
+    svg.append("defs").append("clipPath")
+      .attr("id", "throughputclip2132")
```
julz: why this id? seems pretty random

ui/js/throughput.js
```diff
@@ -0,0 +1,149 @@
...
+    function flattenJSON(data) {
+      var cmds = [];
+      for (var command in data[data.length - 1].Commands) {
+        cmds.push({"cmd": command, "throughput": data[data.length - 1].Commands[command].Throughput})
+      }
+      return cmds;
+    }  
+
+  } //end exports
+  //exports.xAxisMax = function() { return xAxis.scale().domain()[1]; }
```
julz: can we lose the dead code?

ui/js/bar.js
```diff
@@ -200,12 +207,16 @@ var barchart = function(el) {
...
-
-
-
-
+  return {
+    init: function(el){
+      initDOM(el);
+      return exports;
+    },
+    node: d3Graph,
```
julz: why is there both an 'exports' thing which is returned, and also these exported properties.. isn't the point of exports() to be the place where properties are put (unless I'm misreading what's going on here). This seems more complicated than it needs to be.

simonleung8:
the charts now expose function changeState(), used for changing the graph's visibility

julz:
I like the general look, but can we please-please lose the swing-from-the-side animation when the graphs are switched? Just a simple switch without animation would look _way_ better. Also, ideally, I think it'd be great if the tab buttons actually looked like tabs to give the user more idea what's going on, what do you think? (e.g. I achieved the look below by using `border-top-right-radius: 0` and `border-bottom-right-radius: 0`, moving the overall bar 1 pixel in and removing the btn-default class from the inactive button)

cloud_foundry_performance_suite

simonleung8:
Took out the sliding animation, and the buttons look more like tabs now.  Added css binding and tests for 'btn-default' class

julz:
looks better (though, to be honest, Im not sure the fade is really needed; animations should be used to make things clearer, not just to show that we know how to use jquery); the two graphs seem to have different borders which is a bit weird (one graph has a border at the top, the other doesn't).

For the throughput graph, is it supposed to be a bar graph rather than a line chart? On my screen the graph is a horizontal bar, but the icon for the tab (and the nature of the data) suggests it should be a line?..

cloud_foundry_performance_suite

julz:
I just re-read the PR comment and now see it's supposed to be a line "until the data model is ready". It seems like the graph will change quite a lot when the data model is ready. I don't think we can close/merge this story until the data/graph are something more like the final version. 

As of right now if this was merged to master people pulling the latest code would get a half-done implementation that doesn't make much sense and seems like a bug, so let's leave this in the branch until it does something useful.

simonleung8:
PAT is failing because we don't currently save command details into csv files.  PR #64 will fix the problem and travis will pass.

simonleung8:
chart is updated to be a line chart.  
It is very basic for now, some of the features I want the chart to eventually to have:
- throughput value shows when mouse hover over the line
- zoom features

ui/js/tests.js
```diff
@@ -122,22 +148,66 @@ describe("The view", function() {
...
+  it("should draw a line for each command in a workload", function() {
+    var workload = [{ Commands: {
+        "login": {"Throughput": 0.5}, 
+        "push": {"Throughput": 0.1},
+        "list": {"Throughput": 0.3}
+      } }];
+
+    chart(workload);
+
+    expect( d3_throughput.totallines() ).toBe(3);
```
julz: Id rather not use accessors on the object, and instead inspect the generated svg in the DOM. Don't want to have loads of accessors that are just there for the tests, better to check the output matches what we expect

ui/js/tests.js
```diff
@@ -122,22 +148,66 @@ describe("The view", function() {
...
 
-  it("should create a point for each element", function() {
-    chart([1, 2, 3])
-    expect(d3.selectAll('circle').size()).toBe(3)
+})
+
+describe("Throughput chart", function() {  
+  var chart = d3_throughput.init(document.getElementById("target"));
+
+  it("should draw a line for each command in a workload", function() {
```
julz, line 208: there's a lot more logic than this to test. the lines aren't just lines, they're graphing throughput. So we need to set up a throughput graph and assert that the _right_ curves are generated in the svg (goes through correct points in the graph based on the throughput), each workload has a different colour etc. etc. etc.

workloads/workloads.go
```diff
@@ -27,6 +27,7 @@ func DefaultWorkloadList() *WorkloadList {
 		StepWithContext("rest:push", restContext.Push, "Pushes a simple Ruby application using the REST api. This option requires both rest:target and rest:login to be included in the list of workloads"),
 		Step("gcf:push", Push, "Pushes a simple Ruby application using the CF command-line"),
 		Step("dummy", Dummy, "An empty workload that can be used when a CF environment is not available"),
+		Step("dummylogin", Dummy, "An empty workload that can be used when a CF environment is not available"),
```
julz: dummyLogin, but Im getting concerned with how many dummy workloads we have now. If we HAVE to have these, then we should enable them only if an environment variable is specified. But really, let's just start testing against bosh-lite.

simonleung8: There is really no reason to need more 'dummy' functions.  The only purpose for including dummylogin is for feature demonstration.  I will remove the extra 'dummy' once the line chart is accepted.
Anyway, I have improved the tests implementation and coverage

ui/js/throughput.js
```diff
@@ -0,0 +1,186 @@
...
+
+  // Public Properties
+  return {        
+    init: function(el){ //draw DOM elements and return function 'drawGraph' for graph drawing
+      initDOM(el);
+      return drawGraph;
+    },
+    changeState: changeState,
+    node: d3Graph,
+    yAxisMax: function() { return yAxis.scale().domain()[0] },
```
julz: is there a reason anything other than this object needs this property? if not it'd be better not to expose it. If it's just needed by tests it'd be better to directly assert on the result (e.g. the actual svg elements created on the page with the 'axis-label' class, or whatever)

ui/js/tests.js
```diff
@@ -122,22 +148,100 @@ describe("The view", function() {
...
+    var point3 = d3_throughput.scaleX(3) + "," + d3_throughput.scaleY(0.6)
+    expect(paths).toContain(point1);
+    expect(paths).toContain(point2);
+    expect(paths).toContain(point3);
   })
 
-  it("should have a default width (300) and height (500)", function() {
-    expect(chart.width()).toBe(800)
-    expect(chart.height()).toBe(400)
+  it("should draw a line for each command in a workload", function() {
```
julz, line 208: still need to make sure that line goes through the expected points

julz:
as a general minor feature request, would be good to have circular points on the graph associated with each actual data-point.

ui/js/tests.js
simonleung8, line 208: there is a test for that at line 179

simonleung8:
I agree, I actually have started working on this feature, coming up soon

ui/js/tests.js
julz, line 208: my bad, don't know how I missed that

ui/js/tests.js
```diff
@@ -122,22 +148,100 @@ describe("The view", function() {
...
+
+  it("should draw a line to go through points based on the throughput", function() {
+    var workload = [{ Commands: { "login": {"Count": 1, "Throughput": 0.5}}}, 
+                    { Commands: { "login": {"Count": 2, "Throughput": 0.3}}},
+                    { Commands: { "login": {"Count": 3, "Throughput": 0.6}}}];
+
+    chart(workload);
+
+    var paths = $(d3_throughput.node).find("path.line")[0].getAttribute("d")
+    var point1 = d3_throughput.scaleX(1) + "," + d3_throughput.scaleY(0.5)
```
julz: can we do this without exposing scaleX outside of the object?

simonleung8:
improved test implementation

also added circular popup tooltips when mouse hovers over line

simonleung8:
Tests are now verifying the svg elements directly from the DOM, the graphs no longer export any internal state out.

PR #43, opened 2014-02-25
bar chart with error indication

simonleung8:
Barchart will indicate an iteration error in an experiment with a red bar.

Current limitation:
- Hovering over a red bar will trigger a tooltip to show the error message, however, currently the data model does not contain any error message, so the tooltip will always show 'Null' as error.  The limitation should be resolved once the data model design is improved.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -103,6 +103,7 @@ func (ctx *context) handlePush(w http.ResponseWriter, r *http.Request) (interfac
 	worker.AddExperiment("login", experiments.Dummy)
 	worker.AddExperiment("push", experiments.Push)
 	worker.AddExperiment("dummy", experiments.Dummy)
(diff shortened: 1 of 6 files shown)


julz:
is there any way we can just use d3.select(..) on a passed element rather than hardcoding the id everywhere? Don't like using ids if we can help it, it hard-codes an unnecessary assumption that there's only one of each type of graph on a given page. Especially don't like hard-coding 'attr('id', id)' in there, means you have to make sure nothing else on the page uses the same id, and I don't think it's necessary (we can just store the actual graph element rather than selecting on its id).

simonleung8:
All the hardcoded IDs are now gone.  I will update the throughput PR with non-hardcoded ID after this one is merged.

PR #44, opened 2014-03-07
Pat now can be run as a Cloud Foundry APP

simonleung8:
To test:  change 'cf_pat' to be the default branch in settings

Current limitation:
- Real push doesn't work, yet to identify the problem.
- History is lost on server restarts / crashes


julz:
what's the status of this, is it ready to merge? (I know we went back and forth on how it should look in the US, is this the final version?)

simonleung8:
I have rebased this branch, I will rebase the throughput branch once this is merged. Thanks

PR #47, opened 2014-03-17
Implemented keeping track of 95th percentile

jberkhahn:
Implemented support for keeping track of the 95th percentile in the Sample struct. Supports both live and historical data. Updated csv.go to read and write the new field to disk.

### DIFF
--- a/experiment/runner.go
+++ b/experiment/runner.go
@@ -1,6 +1,7 @@
 package experiment
 
 import (
+	"math"
 	"time"
 	. "github.com/cloudfoundry-community/pat/benchmarker"
 )
@@ -34,6 +35,7 @@ type Sample struct {
 	LastResult   time.Duration
 	LastError    error
 	WorstResult  time.Duration
+	NinetyfifthPercentile time.Duration
 	WallTime     time.Duration
 	Type         SampleType
 }
@@ -55,7 +57,7 @@ type ExperimentConfiguration struct {
 type RunnableExperiment struct {
 	ExperimentConfiguration
 	executerFactory func(iterationResults chan IterationResult, errors chan error, workers chan int, quit chan bool) Executable
-	samplerFactory  func(iterationResults chan IterationResult, errors chan error, workers chan int, samples chan *Sample, quit chan bool) Samplable
+	samplerFactory  func(iterations int, iterationResults chan IterationResult, errors chan error, workers chan int, samples chan *Sample, quit chan bool) Samplable
 }
 
 type ExecutableExperiment struct {
@@ -66,6 +68,7 @@ type ExecutableExperiment struct {
 }
 
 type SamplableExperiment struct {
+  maxIterations int
 	iteration chan IterationResult
 	workers   chan int
 	samples   chan *Sample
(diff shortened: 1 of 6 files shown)


experiment/runner_test.go
```diff
@@ -202,6 +204,40 @@ var _ = Describe("ExperimentConfiguration and Sampler", func() {
...
+			go (&SamplableExperiment{maxIterations, iteration, workers, samples, ticks, quit}).Sample()
+		})
+
+		It("Calculates the 95th percentile", func() {
+			for i := 1; i <= maxIterations; i++ {
+				currentIteration := i
+				go func() { iteration <- IterationResult{time.Duration(currentIteration) * time.Second, nil, nil} }()
+			}
+			for q := 1; q <= 20; q++ {
+				Ω((<-samples).NinetyfifthPercentile).Should(Equal(time.Duration(q) * time.Second))
```
julz: I'm trying to figure out why (if) this line works to test the 95th percentile. Could we simplify this and just use pre-determined values: e.g. define two arrays `samplesToSend := []int { 2, 5, 1, 9, 12, 8, 19 }` and `percentileToExpect := []int{ 2, 5, ... };` and then send the first array to the sampler, and expect the outputted samples should have the expected percentiles in order?

experiment/runner.go
```diff
@@ -116,7 +119,7 @@ func (config *RunnableExperiment) Run(tracker func(<-chan *Sample)) error {
 	samples := make(chan *Sample)
 	quit := make(chan bool)
 	done := make(chan bool)
-	sampler := config.samplerFactory(iteration, errors, workers, samples, quit)
+	sampler := config.samplerFactory(config.ExperimentConfiguration.Iterations, iteration, errors, workers, samples, quit)
```
julz: Does this interact correctly with interval and stop etc? You also don't need the `.ExperimentConfiguration.`, Iterations are (I think) accessible directly from `config`

experiment/runner_test.go
```diff
@@ -202,6 +204,39 @@ var _ = Describe("ExperimentConfiguration and Sampler", func() {
...
+			ticks = make(chan int)
+			go (&SamplableExperiment{maxIterations, iteration, workers, samples, ticks, quit}).Sample()
+		})
+
+		It("Calculates the 95th percentile", func() {	
+			samplesToSend := []int { 2, 5, 1, 9, 12, 8, 19, 57, 33, 44, 1, 12, 43, 99, 98, 19, 34, 19, 7, 55, 23}
+			expectedPercentiles := []int { 2, 5, 5, 9, 12, 12, 19, 57, 57, 57, 57, 57, 57, 99, 99, 99, 99, 99, 99, 99, 98}
+			for i := 0; i < maxIterations; i++ {
+				currentIteration := i
+				go func() { iteration <- IterationResult{time.Duration(samplesToSend[currentIteration]) * time.Second, nil, nil} }()
```
julz: I suspect you don't want to spawn maxIteration parallel gofuncs here, you probably want one gofunc with a for loop inside it (otherwise you're relying on them happening to run in order on the channel writes, which isn't guaranteed)

julz:
Just realised this doesn't actually add any display for the new data.. can we add it to cmdline.go at least?

PR #50, opened 2014-03-18
Add list-workloads option

knolleary:


julz:
think you're going to refactor this one and https://github.com/julz/pat/pull/51 at the same time, so going to close this one for now

PR #51, opened 2014-03-18
Reject invalid workload

knolleary:


julz:
What do you think about moving these functions in to Worker? I'd like to reduce the duplication between server.go and cmdline.go, and we're going to need to be able to list and validate workloads in both places. If we push listing (e.g. with a Visit method) and validating into Worker we can share that function and also make it easy to do things like pluggable experiments if we decide we want anything like that.

workloads/workloads_test.go
```diff
@@ -0,0 +1,40 @@
...
+func (self *dummyWorkloadReceiver) AddExperiment(workload Workload) {
+	self.Workloads = append(self.Workloads,workload)
+} 
+
+
+var _ = Describe("Workloads", func() {
+	It("sets the full set of workloads in a worker", func() {
+			
+		testList := []Workload {
+			Workload{"foo", func() error { time.Sleep(1 * time.Second); return nil },"a"},
```
julz: are the `time.Sleep(..)`s doing anything? we can probably lose them

workloads/workloads_test.go
```diff
@@ -0,0 +1,40 @@
...
+		workloadList := WorkloadList{testList}
+		
+		worker := &dummyWorkloadReceiver{}
+		workloadList.DescribeWorkloads(worker)
+
+		for i,w := range testList {
+			Ω(worker.Workloads[i].Name).Should(Equal(w.Name))
+			Ω(worker.Workloads[i].Description).Should(Equal(w.Description))
+		}
+		Ω(len(worker.Workloads)).Should(BeNumerically("==",len(testList)))
```
julz: you can use Ω(worker.Workloads).Should(HaveLen(4)) here

benchmarker/benchmarker.go
```diff
@@ -19,27 +22,28 @@ type IterationResult struct {
 
 type Worker interface {
 	Time(experiment string) IterationResult
-	AddExperiment(name string, fn func() error) Worker
+	AddExperiment(workload Workload)
```
julz: for clarity maybe WorkloadStep rather than Workload (since a -workload flag specifies multiple workload steps)?

benchmarker/benchmarker_test.go
```diff
@@ -34,37 +35,73 @@ var _ = Describe("Benchmarker", func() {
...
+				worker.AddExperiment(Workload{"foo", func() error { time.Sleep(1 * time.Second); return nil },""})
+				ok,err := worker.Validate("foo")
+				Ω(err).Should(BeNil())
+				Ω(ok).Should(BeTrue())
+			})
+			It("Rejects an invalid workload name", func() {
+				worker := NewWorker()
+				worker.AddExperiment(Workload{"foo", func() error { time.Sleep(1 * time.Second); return nil },""})
+				ok,err := worker.Validate("bar")
+				Ω(err).ShouldNot(BeNil())
```
julz, line 93: should probably test the error text contains the invalid field name here

cmdline/cmdline.go
```diff
@@ -19,34 +19,55 @@ type Response struct {
...
 func RunCommandLine() error {
-	lab := NewLaboratory(store.NewCsvStore("output/csvs"))
+	
 	worker := benchmarker.NewWorker()
-	err := RunCommandLineWithLabAndWorker(lab, worker)
+		
+	workloadList.DescribeWorkloads(worker)
+
+	if params.listWorkloads {
+		worker.Visit(printWorkload)
```
julz: can we test this? probably easiest way is to convert `func printWorkload` to `var printWorkload = func() {..}` function and then stub it out in cmdline_test.go

cmdline/cmdline.go
```diff
@@ -19,34 +19,55 @@ type Response struct {
...
+		return nil
+	}
+
+	var ok,err = worker.Validate(params.workload)
```
julz, line 79: needs a test?

julz:
looks good, left a few comments

cmdline/cmdline_test.go
```diff
@@ -15,20 +16,30 @@ var _ = Describe("Cmdline", func() {
...
 	JustBeforeEach(func() {
 		flags = config.NewConfig()
 		InitCommandLineFlags(flags)
 		flags.Parse(args)
-		lab = &dummyLab{}
-		worker = benchmarker.NewWorker()
-		RunCommandLineWithLabAndWorker(lab, worker)
+		LaboratoryFactory = func()(newLab laboratory.Laboratory) {
+			lab = &dummyLab{}
+			newLab = lab
```
julz, line 31: nitpick: just `return lab` would probably be clearer

knolleary, line 31: That is where I started, but it causes: `runtime error: invalid memory address or nil pointer dereference`

This, admittedly inelegant, variable swap was the only way I could find to get it to work.

cmdline/cmdline.go
```diff
@@ -19,45 +19,74 @@ type Response struct {
...
+		
+	if params.listWorkloads {
+		worker.Visit(PrintWorkload)
+		return nil
+	}
+
+	var ok,err = worker.Validate(params.workload)
+	
+	if !ok {
+		fmt.Printf("Unrecognised workload: '%s'\n\n",err)
```
julz: this might be assuming more than it should about what the error message contains? Maybe 'Invalid workload: '?

PR #59, opened 2014-03-25
Force push reuploads

jberkhahn:
Added GenerateAndPush experiment; forces reupload via modifying files to change app SHA
Copies the app to be pushed (currently hardcoded to the default hello_world app) into a temporary directory in assets. Cats a randomly generated string to a comment at the end of each file, pushes the app, and then deletes the directory.


julz:
travis seems to be barfing on that commit, can you try to get it running?

workloads/gcf.go
```diff
@@ -34,3 +37,35 @@ func Push() error {
...
 }
+
+func MoveAndSalt(srcDir string, dstDir string, salt string){
+	os.Mkdir(dstDir, 0777)
+	files, _ := ioutil.ReadDir(srcDir)	          
+	for i := 0; i < len(files); i++ {
+		fileInfo, _ := os.Stat(srcDir + files[i].Name())
+		if fileInfo.Mode().IsDir() {
+			newDstDir := dstDir + "/" + fileInfo.Name()
+			MoveAndSalt(srcDir + fileInfo.Name()+"/", newDstDir+"/", salt)
```
julz: [path.Join](http://golang.org/pkg/path/#Join) would be better, slashes aren't cross-platform

workloads/gcf.go
```diff
@@ -34,3 +37,35 @@ func Push() error {
 	err := Cf("push", "pats-"+guid.String(), "patsapp", "-m", "64M", "-p", "assets/hello-world").ExpectOutput("App started")
 	return err
 }
+
+func MoveAndSalt(srcDir string, dstDir string, salt string){
+	os.Mkdir(dstDir, 0777)
+	files, _ := ioutil.ReadDir(srcDir)	          
+	for i := 0; i < len(files); i++ {
```
julz: How about letting something like [filepath.Walk](http://golang.org/pkg/path/filepath/#Walk) handle the traversal, might simplify this function down a bit

workloads/gcf.go
```diff
@@ -34,3 +37,35 @@ func Push() error {
...
+		fileInfo, _ := os.Stat(srcDir + files[i].Name())
+		if fileInfo.Mode().IsDir() {
+			newDstDir := dstDir + "/" + fileInfo.Name()
+			MoveAndSalt(srcDir + fileInfo.Name()+"/", newDstDir+"/", salt)
+		}
+		if fileInfo.Mode().IsRegular() {
+			input, _ := ioutil.ReadFile(srcDir + files[i].Name())
+			output, _ := os.Create(dstDir+"/"+files[i].Name())
+			output.Write(input)
+			output.Write([]byte("\n#" + salt))
```
julz: Should we be slightly cleverer here and e.g. only replace $RANDOM_TEXT in files in the src directory with the random string? That way we could make sure the text was put somewhere valid (e.g. in a comment in a src file etc.) where it won't stop the app starting (which might be required by later workloads that want to actually do something with the started app).

workloads/gcf.go
```diff
@@ -34,3 +37,35 @@ func Push() error {
...
+		}        
+	}
+}  
+
+func GenerateAndPush() error {
+	srcDir := "assets/hello-world/"
+	guid, _ := uuid.NewV4()
+	rand.Seed(time.Now().UTC().UnixNano())
+	salt := strconv.FormatInt(rand.Int63(), 10)
+	dstDir := "assets/"+salt
```
julz: should really put this in a [tmp directory](http://golang.org/pkg/os/#TempDir) and clean it up after you're done; these could be big apps getting copied hundreds of times in an experiment

workloads/gcf_test.go
```diff
@@ -0,0 +1,48 @@
...
+		dstDir = "../assets/"+salt + "/"
+	})
+
+	Describe("Generating and Pushing an app", func() {
+		Context("MoveAndSalt", func() {
+			It("Creates a new directory and adds a comment to each file to change the hash", func() {
+				MoveAndSalt(srcDir, dstDir, salt)
+				
+				files, _ := ioutil.ReadDir(srcDir)
+				for i := 0; i < len(files); i++ {
```
julz: should avoid duplicating the function code in the test; it's not much of a test that way. Better to set up a test directory you know the structure of, and then assert against the files within it explicitly rather than duplicating the recursion code. 

e.g. if you know the test directory has a file `foo/bar/baz`, just check that `$destdir/foo/bar/baz` has a modified hash -- no need to recurse down.

workloads/workloads.go
```diff
@@ -26,6 +26,7 @@ func DefaultWorkloadList() *WorkloadList {
 		StepWithContext("rest:login", restContext.Login, "Performs a login to the REST api. This option requires rest:target to be included in the list of workloads"),
 		StepWithContext("rest:push", restContext.Push, "Pushes a simple Ruby application using the REST api. This option requires both rest:target and rest:login to be included in the list of workloads"),
 		Step("gcf:push", Push, "Pushes a simple Ruby application using the CF command-line"),
+		Step("gcf:generateandpush", GenerateAndPush, "Generates and Pushes a unique simple Ruby application using the CF command-line"),
```
julz: for consistency this should probably be generateAndPush (camel case), or we can make it gcf:generate-and-push and update dummyWithErrors to dummy-with-errors. Either way is fine but they should match.

julz:
left a boatload of small comments, hopefully travis will [unbreak its heart](https://www.youtube.com/watch?feature=player_detailpage&v=pU2LzuVrqLQ#t=81) in the process of addressing them

workloads/gcf.go
```diff
@@ -34,3 +40,35 @@ func Push() error {
 	err := Cf("push", "pats-"+guid.String(), "patsapp", "-m", "64M", "-p", "assets/hello-world").ExpectOutput("App started")
 	return err
 }
+
+func MoveAndSalt(srcDir string, dstDir string, salt string){
+	filepath.Walk(srcDir,  func(file string, info os.FileInfo, err error) error { 
+		if err != nil { panic(err) }
```
julz: panic is a bit extreme, just make sure the workload returns an err so it gets reported

workloads/gcf_test.go
```diff
@@ -0,0 +1,53 @@
...
+			
+				oldSub, _ := ioutil.ReadFile(path.Join(srcDir, "subdir", "subfile.txt"))
+				newSub, err := ioutil.ReadFile(path.Join(dstDir, "subdir", "subfile.txt"))
+				oldSubHash := md5.Sum(oldSub)
+				newSubHash := md5.Sum(newSub)
+				
+				Ω(newSubHash).ShouldNot(Equal(oldSubHash))
+				Ω(err).ShouldNot(HaveOccured())
+				
+				os.RemoveAll(dstDir)
```
julz: this should go in to an AfterEach so it happens even if one of the above steps fails

workloads/gcf.go
```diff
@@ -34,3 +40,35 @@ func Push() error {
...
+		pathTail := strings.SplitAfter(file, srcDir)[1]
+		if info.IsDir() {
+			os.Mkdir(path.Join(dstDir, pathTail), 0777)
+		} else if info.Mode().IsRegular() {
+			input, _ := ioutil.ReadFile(file)
+			inputString := strings.Replace(string(input), "$RANDOM_TEXT", salt, -1)
+			input = []byte(inputString)
+			output, _ := os.Create(path.Join(dstDir, pathTail))
+			output.Write(input)
+			//output.Write([]byte("\n#" + salt))
```
julz: lose dead code

workloads/gcf.go
```diff
@@ -34,3 +40,35 @@ func Push() error {
...
+			output.Close()
```
julz: should use `defer output.Close()` right after `os.Create(..)` (so the file is closed even if e.g. Write fails etc.)

workloads/gcf.go
```diff
@@ -34,3 +40,35 @@ func Push() error {
...
 }
+
+func MoveAndSalt(srcDir string, dstDir string, salt string){
+	filepath.Walk(srcDir,  func(file string, info os.FileInfo, err error) error { 
+		if err != nil { panic(err) }
+		pathTail := strings.SplitAfter(file, srcDir)[1]
+		if info.IsDir() {
+			os.Mkdir(path.Join(dstDir, pathTail), 0777)
+		} else if info.Mode().IsRegular() {
+			input, _ := ioutil.ReadFile(file)
```
julz: for both this line, and L54, better to return the error instead of hiding it, then the workload can fail with a helpful reason

workloads/gcf.go
```diff
@@ -34,3 +40,35 @@ func Push() error {
...
+	})
+} 
+
+func GenerateAndPush() error {
+	guid, _ := uuid.NewV4()
+	srcDir := path.Join("assets", "hello-world")
+	rand.Seed(time.Now().UTC().UnixNano())
+	salt := strconv.FormatInt(rand.Int63(), 10)
+	dstDir := path.Join(os.TempDir(), salt)
+	MoveAndSalt(srcDir, dstDir, salt)
```
julz: maybe rename `MoveAndSalt` -> `CopyAndReplaceText(srcDir, dstDir, searchText, replaceText` since that's what it does (it neither moves, nor salts); then we can move the text generation in to a simple function called e.g. `randomSingleLineText` or something

workloads/gcf_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	. "github.com/cloudfoundry-community/pat/workloads"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("GCF Workloads", func() {
+	Describe("Generating and Pushing an app", func() {
+		Context("MoveAndSalt", func() {
+			It("Creates a new directory and adds a comment to each file to change the hash", func() {
```
julz: would be clearer to have these tests each test one simple thing rather than effectively a big integration test, e.g. "When there are directories: It copies the directories", "When there are files: It copies the files", "When there are files: It copies the files: And when they contain the searchText: It replaces the searchText". Coding test-first would encourage this style. (the big advantage of this style is when a test fails it's clear which bit broke, and the coverage is normally better).

workloads/gcf.go
```diff
@@ -34,3 +40,35 @@ func Push() error {
...
+
+	err := Cf("push", "pats-"+guid.String(), "patsapp", "-m", "64M", "-p", dstDir).ExpectOutput("App started")
+	os.RemoveAll(dstDir)
```
julz: probably `defer os.RemoveAll(dstDir)` just after the create is clearer

PR #60, opened 2014-03-25
Simplified throughput calculation

simonleung8:
Currently throughput is being calculated every second, however, there is no point in doing so because command.TotalTime only gets populated at the end of each iteration.

To fix this, throughput calculation is now done at each iteration, this implementation eliminates the extra channel in SamplableExperiment{} and is much cleaner

Also added a heatbeat timer in Sample() so users could see the Wall Time in CLI runs, without heatbeat users only see updates at the end of each iteration

### DIFF
--- a/experiment/runner_test.go
+++ b/experiment/runner_test.go
@@ -144,7 +144,6 @@ var _ = Describe("ExperimentConfiguration and Sampler", func() {
 			iteration chan IterationResult
 			workers   chan int
 			quit      chan bool
-			ticks     chan int
 			samples   chan *Sample
 		)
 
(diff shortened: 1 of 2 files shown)


experiment/runner_test.go
```diff
@@ -182,24 +175,14 @@ var _ = Describe("ExperimentConfiguration and Sampler", func() {
...
-
-			<-samples // ResultSample
-			<-samples // ResultSample
-			Ω((<-samples).Commands["push"].Throughput).Should(BeNumerically("==", 1))
-			<-samples // ResultSample
+				
 			Ω((<-samples).Commands["push"].Throughput).Should(BeNumerically("==", 1))
-			<-samples // ResultSample
-			Ω((<-samples).Commands["push"].Throughput).Should(BeNumerically("==", 4.0/6.0))
+			Ω((<-samples).Commands["list"].Throughput).Should(BeNumerically("==", 0.5))
```
julz: can we make this test a bit more convincing, as it is it only shows the case where there's one operation (list or push) and doesn't show what happens if a workload runs a particular operation more than once, or if the command is used over multiple iterations (Etc).

simonleung8:
Improved test case

PR #61, opened 2014-03-26
Redis-based worker implementation

julz:
This is most of the code needed for redis-based distributed worker task sharing. There are still a couple edges here: for example whichever instance triggers the job is a single-point-of-failure because it must continue to monitor the job. This doesn't seem like a huge deal for the purposes of this sort of project. Additionally right now only the triggering instance can see a running experiment in the UI because the others do not refresh state from the store except on startup, will fix that next and call this story done.


julz:
Updated so any server can view a running experiment 'live". In the process dropped the buffering from Laboratory, finally. Think this completes this feature.

simonleung8:
The bugs we have seen earlier are now fixed. Merging

PR #63, opened 2014-03-30
Proper output file close

simonleung8:
Fix the bug where file might not be closed properly and cause ginkgo tests to fail


julz:
sorry to be a pedant, but can we lose the "bug fix:" prefixes from the commit messages? apart from that looks good

simonleung8:
no problem.  commit message changed

PR #64, opened 2014-03-31
Saving commands into cvs file

simonleung8:
With this changes, commands map will be persisted in csv files along with all the iteration details

### DIFF
--- a/store/csv.go
+++ b/store/csv.go
@@ -54,12 +54,27 @@ func (self *csvFile) Write(samples <-chan *experiment.Sample) {
 	}
 	defer f.Close()
 
+	var header []string
+	var body []string
 	w := csv.NewWriter(f)
-	w.Write([]string{"Average", "TotalTime", "Total", "TotalErrors", "TotalWorkers", "LastResult", "WorstResult", "NinetyfifthPercentile", "WallTime", "Type"})
 
 	for s := range samples {
 		if s.Type == experiment.ResultSample {
-			w.Write([]string{strconv.Itoa(int(s.Average.Nanoseconds())),
+
+			if len(header) == 0 {
+				header = []string{"Average", "TotalTime", "Total", "TotalErrors", "TotalWorkers", "LastResult", "WorstResult", "NinetyfifthPercentile", "WallTime", "Type"}
+				for k, _ := range s.Commands {
+					header = append(header, "Commands:"+k+":Count",
+						"Commands:"+k+":Throughput",
+						"Commands:"+k+":Average",
+						"Commands:"+k+":TotalTime",
+						"Commands:"+k+":LastTime",
+						"Commands:"+k+":WorstTime")
+				}
+				w.Write(header)
+			}
+
+			body = []string{strconv.Itoa(int(s.Average.Nanoseconds())),
 				strconv.Itoa(int(s.TotalTime.Nanoseconds())),
 				strconv.Itoa(int(s.Total)),
 				strconv.Itoa(int(s.TotalErrors)),
@@ -68,7 +83,18 @@ func (self *csvFile) Write(samples <-chan *experiment.Sample) {
 				strconv.Itoa(int(s.WorstResult.Nanoseconds())),
 				strconv.Itoa(int(s.NinetyfifthPercentile.Nanoseconds())),
 				strconv.Itoa(int(s.WallTime)),
-				strconv.Itoa(int(s.Type))})
+				strconv.Itoa(int(s.Type))}
+
+			for k, _ := range s.Commands {
+				body = append(body, strconv.Itoa(int(s.Commands[k].Count)),
+					strconv.FormatFloat(s.Commands[k].Throughput, 'f', 8, 64),
+					strconv.Itoa(int(s.Commands[k].Average.Nanoseconds())),
+					strconv.Itoa(int(s.Commands[k].TotalTime.Nanoseconds())),
+					strconv.Itoa(int(s.Commands[k].LastTime.Nanoseconds())),
+					strconv.Itoa(int(s.Commands[k].WorstTime.Nanoseconds())))
+			}
+
+			w.Write(body)
 			w.Flush()
 		}
 	}
--- a/store/csv.go
+++ b/store/csv.go
@@ -112,10 +112,18 @@ func (self *csvFile) GetData() (samples []*experiment.Sample, err error) {
 		return nil, err
 	}
 
+	var cmd experiment.Command
+	var keys = make(map[string]bool)
 	for i, d := range decoded {
 		if i == 0 {
+			for _, s := range d {
+				if strings.HasPrefix(s, "Commands:") {
+					keys[strings.Split(s, ":")[1]] = true
+				}
+			}
 		} else {
 			sample := &experiment.Sample{}
+			sample.Commands = make(map[string]experiment.Command)
 			sample.Average, err = duration(d[0])
 			sample.TotalTime, err = duration(d[1])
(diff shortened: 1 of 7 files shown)


store/csv_test.go
```diff
@@ -33,9 +34,12 @@ var _ = Describe("Csv Store", func() {
 			os.RemoveAll(dir)
 			store = NewCsvStore(dir)
 			writer := store.Writer("foo")
+			commands = make(map[string]experiment.Command)
+			cmd := experiment.Command{1, 1, 1, 1, 1, 1}
```
julz: lets make each of these numbers different to make sure we round-tripped the columns in the right order

store/csv.go
```diff
@@ -54,12 +54,27 @@ func (self *csvFile) Write(samples <-chan *experiment.Sample) {
...
 	w := csv.NewWriter(f)
-	w.Write([]string{"Average", "TotalTime", "Total", "TotalErrors", "TotalWorkers", "LastResult", "WorstResult", "NinetyfifthPercentile", "WallTime", "Type"})
 
 	for s := range samples {
 		if s.Type == experiment.ResultSample {
-			w.Write([]string{strconv.Itoa(int(s.Average.Nanoseconds())),
+
+			if len(header) == 0 {
+				header = []string{"Average", "TotalTime", "Total", "TotalErrors", "TotalWorkers", "LastResult", "WorstResult", "NinetyfifthPercentile", "WallTime", "Type"}
+				for k, _ := range s.Commands {
```
julz: hmmm... does the first sample always contain all the commands that are going to be used? Maybe we need to pass the list of commands to store as columns to the csv.go constructor?

simonleung8:
csv now gets a full list commands from workloads; also changed the delimiter from ":" to "|", since we use ":" in our commands.  And test numbers are updated

store/csv.go
```diff
@@ -54,12 +62,27 @@ func (self *csvFile) Write(samples <-chan *experiment.Sample) {
 	}
 	defer f.Close()
 
+	var header []string
+	var body []string
 	w := csv.NewWriter(f)
-	w.Write([]string{"Average", "TotalTime", "Total", "TotalErrors", "TotalWorkers", "LastResult", "WorstResult", "NinetyfifthPercentile", "WallTime", "Type"})
+
+	if len(header) == 0 {
```
julz: when is len(header) not 0?

store/csv.go
```diff
@@ -32,11 +34,17 @@ func (store *CsvStore) Writer(guid string) func(samples <-chan *experiment.Sampl
...
 
 func (store *CsvStore) load(filename string, guid string) (experiment.Experiment, error) {
-	return &csvFile{path.Join(store.dir, filename), guid}, nil
+	return &csvFile{path.Join(store.dir, filename), guid, nil}, nil
 }
 
 func (store *CsvStore) newCsvFile(guid string) *csvFile {
-	return &csvFile{path.Join(store.dir, strconv.Itoa(int(time.Now().UnixNano()))+"-"+guid+".csv"), guid}
+	file := &csvFile{path.Join(store.dir, strconv.Itoa(int(time.Now().UnixNano()))+"-"+guid+".csv"), guid, nil}
+	workloads.DefaultWorkloadList().DescribeWorkloads(file)
```
julz: would be better to inject a workload list (i.e. pass it into the constructor), that way the tests don't have to assume a 'dummy' workload exists and we're not tightly coupled against the singleton DefaultWorkloadList inside csvstore (I'm ok with files called config.go knowing about DefaultWorkloadList, but others shouldn't, ideally).

store/csv.go
simonleung8: you are right, header length will always be 0 since I moved header outside of the loop in the new code. will update

simonleung8:
worklist is now injected from config, and tests are getting commands from workloads dynamically

store/csv_test.go
```diff
@@ -23,19 +30,25 @@ var _ = Describe("Csv Store", func() {
...
+			store    *CsvStore
+			output   string
+			commands map[string]experiment.Command
 		)
 
 		JustBeforeEach(func() {
 			dir = "/var/tmp/test-output/csvstore"
 			os.RemoveAll(dir)
-			store = NewCsvStore(dir)
+			store = NewCsvStore(dir, workloads.DefaultWorkloadList())
```
julz: no need to use DefaultWorkloadList() at all here, and by using it you make the tests fail if we change the contents of the default workload list. You can just create your own workload list instance and pass that, e.g. like this: https://github.com/cloudfoundry-community/pat/blob/9c500c9c7ab84f8e5b93dbef34f41c2ba95a18fb/workloads/workloads_test.go#L17

store/csv_test.go
```diff
@@ -10,10 +10,17 @@ import (
 
 	"github.com/cloudfoundry-community/pat/experiment"
 	. "github.com/cloudfoundry-community/pat/store"
+	"github.com/cloudfoundry-community/pat/workloads"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
+type dummyWorkloadReceiver struct{ Workloads []string }
```
julz: the tests shouldn't need to receive any workloads, so you can lose this struct if you create your own workloadlist instead of passing the default one (see below comment)

store/csv.go
```diff
@@ -86,10 +123,18 @@ func (self *csvFile) GetData() (samples []*experiment.Sample, err error) {
 		return nil, err
 	}
 
+	var cmd experiment.Command
+	var keys = make(map[string]bool)
 	for i, d := range decoded {
 		if i == 0 {
+			for _, s := range d {
+				if strings.HasPrefix(s, "Commands|") {
+					keys[strings.Split(s, "|")[1]] = true
```
julz: why is this a map?? you only ever assign true to any keys. The thing is, a map doesn't guarantee order, so the commands can be parsed in any order. So when we come to get the value for a particular command below, we'll - most of the time - be getting the value of the wrong command. Surely this should just be a []string?

store/csv.go
```diff
@@ -101,6 +146,22 @@ func (self *csvFile) GetData() (samples []*experiment.Sample, err error) {
 			sample.WallTime, err = duration(d[8])
 			sample.Type = experiment.ResultSample // this is the only type we currently persist
 
+			var i = 10
```
julz: not only is this a magic number that has to exactly match the number of previous fields, it's shadowing the `i` variable from the enclosing range statement. There's really no good reason to re-use `i` here and it's super-confusing.

store/csv.go
julz: or, even better, make it a `map[string]int` (command name -> column number), and then you can lose the `var i` below and just directly use the column index from the header. Also why is it called 'keys' rather than something like 'commands' or 'commandColumns'?

simonleung8:
improved implementation, testing is now fully independent from the default workload list.
also a map[string]int is now used to keep track of column orders for reading from csv files

store/csv_test.go
```diff
@@ -10,10 +10,15 @@ import (
 
 	"github.com/cloudfoundry-community/pat/experiment"
 	. "github.com/cloudfoundry-community/pat/store"
+	"github.com/cloudfoundry-community/pat/workloads"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
+var testList []workloads.WorkloadStep = []workloads.WorkloadStep{
```
julz: Any reason this is up here rather than inline below?

simonleung8: no particular reason, just updated for it to be inline

PR #68, opened 2014-04-07
small update to address firefox bug

simonleung8:
- There is a bug in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=612118) causing the chart to crash if SVG element is not rendered (display=none).  I have added a try..catch to get around the bug.
- Also gave the line chart a little headroom in the y-axis for nicer rendering
- changed svg text 'dx' to 'x' for Firefox rendering, tested in Firefox, Chrome and Safari
- popup tooltip in line chart is colored the same as the line now
- reduced tooltip size

### DIFF
--- a/ui/js/bar.js
+++ b/ui/js/bar.js
@@ -96,6 +96,8 @@ d3_workload = function() {
   } //end initDOM
 
   var drawGraph = function(data) {
+    if (!data[0]) return;
+
     const second = 1000000000;
     var len = data.length;
     x = x.domain( [1, len, 1] ).range([barWidth + 1, len * (barWidth + 1)]);
@@ -193,7 +195,14 @@ d3_workload = function() {
   }
 
   function getNodeWidth(node) {
-    return node.getBBox().width;
+    //using a try..catch block to get around the Firefox bug, getBBox() fails if svg element is not rendered (display = none)
+    //https://bugzilla.mozilla.org/show_bug.cgi?id=612118
+    try {
+      return node.getBBox().width;
+    }
+    catch(err) {
+      return 0
+    }
   }
 
   function transformChart(node, x, y) {
--- a/ui/js/throughput.js
+++ b/ui/js/throughput.js
@@ -1,7 +1,7 @@
 d3_throughput = function() {
   var margin = {top: 50, right: 30, bottom: 30, left: 30};  
   var svgWidth, svgHeight, jqObj, d3Obj, drawArea;
-  var x, y, xAxis, yAxis, svg, graphBox;
+  var x, y, xAxis, yAxis, svg, graphBox, color;
 
   var d3Graph = document.createElement('div');
   d3Graph.className = "throughputContainer";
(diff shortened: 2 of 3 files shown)


ui/js/throughput.js
```diff
@@ -164,19 +164,19 @@ d3_throughput = function() {
...
       .append("g")
       .attr("class", "data" + d.cmd)
     tooptip.append("circle")        
-        .style("fill", "#333")
+        .style("fill", c)
         .attr("class", d.cmd)
         .attr("cx", function(d, i){ return x(i) })
         .attr("cy", function(d){ return y(d) })
-        .attr("r", function(d, i) { return i?15:0 })
+        .attr("r", function(d, i) { return i?13:0 })
```
julz: make the sizes configurable rather than a magic number, and add a test

ui/js/throughput.js
```diff
@@ -164,19 +164,19 @@ d3_throughput = function() {
...
   } //end drawGraph
 
-  function drawToolTip(d) {
+  function drawToolTip(d, c) {
     var tooptip = svg.selectAll("g.data" + d.cmd).data(d.throughput).enter()
       .append("g")
       .attr("class", "data" + d.cmd)
     tooptip.append("circle")        
-        .style("fill", "#333")
+        .style("fill", c)
```
julz, line 174: test

ui/js/throughput.js
```diff
@@ -113,7 +113,7 @@ d3_throughput = function() {
         .attr("class", "line")
         .attr("d", function(d, i) {return line(d.throughput); })
         .style("stroke", function (d) { return color(d.cmd) })  
-        .on("mouseover", function(d){ drawToolTip(d) })      
+        .on("mouseover", function(d){ drawToolTip(d, color(d.cmd)) })
```
julz, line 116: test..

ui/js/bar.js
```diff
@@ -193,7 +195,12 @@ d3_workload = function() {
   }
 
   function getNodeWidth(node) {
-    return node.getBBox().width;
+    try {
```
julz, line 200: this needs a test. If you absolutely totally can't find a way to test it, then at least put a comment in referencing the mozilla bug so someone knows later why this code is in there.

ui/js/bar.js
```diff
@@ -96,6 +96,8 @@ d3_workload = function() {
   } //end initDOM
 
   var drawGraph = function(data) {
+    if (!data[0]) return;
```
julz, line 99: test..

ui/js/throughput.js
simonleung8: Just saw Andy's comment, he actually wants the circle to be tiny with the text offset from it.  I have updated the tooltip to what he want.  
Other tests are updated.

julz: tests for other bits look ok.

this would still be much better if the tooltip size was configurable (i.e. pass to constructor with default) and tested, or at LEAST exported to a constant rather than being arbitrarily hardcoded to 13, it's good to [avoid mysterious numbers](http://c2.com/cgi/wiki?MagicNumber) in the code because it's all too easy to forget where to change things later. this line would be much better if it read `.attr("r", function(d,i) { return self.pointRadius }`or something (btw what's the idea of not showing if `i != true`?)

simonleung8: You are right about the magic number.  I have made it a const instead of making it configurable.  I don't see the need to make it configurable at the moment.

Also took away "i != true", I was thinking no point of showing the throughput of 0 iteration, but showing it might give users a more complete data diagram.

PR #70, opened 2014-04-09
remove tooltip illegal character

simonleung8:
Currently throughput tooltip uses command name as class name, and it might contain illegal characters (e.g. gcf:push).

This change make sure illegal character will not end up in class name
no test provided because I don't think it is necessary to test logic within private methods

### DIFF
--- a/ui/js/tests.js
+++ b/ui/js/tests.js
@@ -254,6 +254,18 @@ describe("Throughput chart", function() {
     expect( $(node).find("g.datalogin circle")[0].getAttribute("fill") ).toEqual(color)
   })
 
+  it("should replace illegal characters with underscore in tooltip class names", function() {
+    var illegalName = "gcf:login+123";
+    var workload = [{ Commands: { illegalName: {"Count": 1, "Throughput": 0.50}}}];
+
+    chart(workload);
+    d3.select(node).select("path.line").on("mouseover")({cmd: illegalName, throughput:[0.50]});
+
+    expect( $(node).find("g.data" + illegalName).length ).toEqual(0);
+        
+    expect( $(node).find("g.data" + "gcf_login_123").length ).toEqual(1);
+  })  
+
   it("should show the maximum command throughput in seconds in the y-axis", function() {
     var workload = [{ Commands: {
         "login": {"Throughput": 0.5}, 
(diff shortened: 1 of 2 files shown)


julz:
It's necessary to test the public effect. For example that tooltips are generated with a valid classname even when an experiment has an illegal character.

simonleung8:
You are right.  Backfilled the fix with test

ui/js/tests.js
```diff
@@ -254,6 +254,19 @@ describe("Throughput chart", function() {
     expect( $(node).find("g.datalogin circle")[0].getAttribute("fill") ).toEqual(color)
   })
 
+  it("it should remove illegal characters in tooltips class name", function() {
```
julz: are there other illegal chars we should remove while we're thinking about this?

ui/js/tests.js
```diff
@@ -254,6 +254,19 @@ describe("Throughput chart", function() {
...
+    var illegalName = "gcf:login";
+    var workload = [{ Commands: { illegalName: {"Count": 1, "Throughput": 0.50}}}];
+
+    chart(workload);
+    d3.select(node).select("path.line").on("mouseover")({cmd: illegalName, throughput:[0.50]});
+
+    expect( $(node).find("g.data" + illegalName).length ).toEqual(0);
+    
+    var legalName = illegalName.replace(":","");
```
julz: why repeat the logic in the test here? you know that the legal version of the name is "g.datagcflogin"

simonleung8: I was thinking it is easier to read, anyway, took out the repeating logic

ui/js/tests.js
simonleung8: Now it checks for all non A..Z 0..9 chars, and replace them with "_"

ui/js/throughput.js
```diff
@@ -167,12 +167,12 @@ d3_throughput = function() {
   function drawToolTip(d, c) {
     const pointRadis = 13;
     const yOffset = 4;
-    var tooptip = svg.selectAll("g.data" + d.cmd).data(d.throughput).enter()
+    var tooptip = svg.selectAll("g.data" + d.cmd.replace(/[^a-zA-Z0-9]/g,"_")).data(d.throughput).enter()
```
julz: lets remove the duplication of the regex? otherwise, looks ok

simonleung8:
No more duplication of regex.

PR #73, opened 2014-04-12
Allow multiple credentials

simonleung8:
This modification allows multiple username and password in the -rest:username and -rest:password flags.
All username and password must be separated by commas. PAT take each username and password in turn at each iteration, and loop back when the list ends

Example usage
go run main.go -concurrency=1 -iterations=2 -rest:target=http://api.10.244.0.34.xip.io -rest:space=test -rest:username=admin,admin2 -rest:password=admin,pass2 -workload=rest:target,rest:login

### DIFF
--- a/benchmarker/local.go
+++ b/benchmarker/local.go
@@ -15,10 +15,11 @@ func NewLocalWorker() *LocalWorker {
 	return &LocalWorker{defaultWorker{make(map[string]workloads.WorkloadStep)}}
 }
 
-func (self *LocalWorker) Time(experiment string) (result IterationResult) {
+func (self *LocalWorker) Time(experiment string, workerIndex int) (result IterationResult) {	
 	experiments := strings.Split(experiment, ",")
 	var start = time.Now()
 	context := make(map[string]interface{})
+	context["workerIndex"] = workerIndex
 	for _, e := range experiments {
 		stepTime, err := Time(func() error { return self.Experiments[e].Fn(context) })
 		result.Steps = append(result.Steps, StepResult{e, stepTime})
--- a/benchmarker/redis.go
+++ b/benchmarker/redis.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"strconv"
 
 	"github.com/cloudfoundry-community/pat/redis"
 	"github.com/cloudfoundry-community/pat/workloads"
@@ -26,12 +27,13 @@ func NewRedisWorkerWithTimeout(conn redis.Conn, timeoutInSeconds int) Worker {
 	return &rw{defaultWorker{make(map[string]workloads.WorkloadStep)}, conn, timeoutInSeconds}
 }
 
-func (rw rw) Time(experiment string) (result IterationResult) {
+func (rw rw) Time(experiment string, workerIndex int) (result IterationResult) {
 	guid, _ := uuid.NewV4()
-	rw.conn.Do("RPUSH", "tasks", "replies-"+guid.String()+" "+experiment)
+	rw.conn.Do("RPUSH", "tasks", "replies-"+guid.String()+" "+strconv.Itoa(workerIndex)+" "+experiment)	
 	reply, err := redis.Strings(rw.conn.Do("BLPOP", "replies-"+guid.String(), rw.timeoutInSeconds))
 
 	if err != nil {
+		fmt.Println(encodeError(err))
 		return IterationResult{0, []StepResult{}, encodeError(err)}
 	} else {
 		json.Unmarshal([]byte(reply[1]), &result)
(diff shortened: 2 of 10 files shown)


bravera:
I am out of the office until 22/04/2014.

For urgent issues regarding the IBM CloudFoundry Community Development team
please contact Alex Tarpinian/Austin/IBM.

For other issues please contact my manager, David Watson/UK/IBM.

Note: This is an automated response to your message  "[pat] Allow multiple
credentials (#73)" sent on 12/04/2014 21:28:00.

This is the only notification you will receive while this person is away.

benchmarker/local.go
```diff
@@ -15,10 +15,11 @@ func NewLocalWorker() *LocalWorker {
 	return &LocalWorker{defaultWorker{make(map[string]workloads.WorkloadStep)}}
 }
 
-func (self *LocalWorker) Time(experiment string) (result IterationResult) {
+func (self *LocalWorker) Time(experiment string, workerIndex int) (result IterationResult) {	
 	experiments := strings.Split(experiment, ",")
 	var start = time.Now()
 	context := make(map[string]interface{})
+	context["workerId"] = workerIndex
```
julz: let's call it worker.index instead of workerId (since we rely on the property that it's a sequentially increasing index and would have problems if it was any old id). Note you already call it workerIndex in the parameter list, so let's keep that consistent in the context.

workloads/rest.go
```diff
@@ -51,8 +52,9 @@ func (r *rest) Target(ctx map[string]interface{}) error {
 
 func (r *rest) Login(ctx map[string]interface{}) error {
 	body := &LoginResponse{}
+	workerId, _ := ctx["workerId"].(int)
```
julz: I actually think it'd be better to calculate the right user/password here inside Login (perhaps by calling out to a function like e.g. `credentialsForWorker(workerID) string, string`), and pass them to oauthInputs, that way oauthInputs is a generic function taking a username/password and returning a proper oauthInput message rather than being tied to this particular way of getting usernames and password for a workload.

julz:
this looks good apart from the minor comments

simonleung8:
@julz: Updated with the changes you recommended

benchmarker/redis.go
```diff
@@ -77,7 +77,7 @@ func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
 		if err == nil {
 			parts := strings.SplitN(reply[1], " ", 2)
 			go func(experiment string, replyTo string) {
-				result := delegate.Time(experiment)
+				result := delegate.Time(experiment, 1)
```
julz: we can do better here; is there a good reason not to send the actual workerId over to the redis worker?

julz:
looks good, let's make sure to pass the workerIndex over the wire to the remote redis workers and then I think this is gtg

simonleung8:
workerIndex is now passed to remote workers.

benchmarker/redis.go
```diff
@@ -26,9 +27,9 @@ func NewRedisWorkerWithTimeout(conn redis.Conn, timeoutInSeconds int) Worker {
 	return &rw{defaultWorker{make(map[string]workloads.WorkloadStep)}, conn, timeoutInSeconds}
 }
 
-func (rw rw) Time(experiment string) (result IterationResult) {
+func (rw rw) Time(experiment string, workerIndex int) (result IterationResult) {
 	guid, _ := uuid.NewV4()
-	rw.conn.Do("RPUSH", "tasks", "replies-"+guid.String()+" "+experiment)
+	rw.conn.Do("RPUSH", "tasks", "replies-"+guid.String()+" "+experiment+" "+strconv.Itoa(workerIndex))
```
julz: this relies pretty heavily on the experiment string not containing spaces. What happens if someone specifies the experiment string as "login, push, push"? Can that get this far down? I'm not sure it can right now, but things change and we could stand to be more careful about it here: either we should catch the error higher up or sanitise any spaces out of the string before sending them over. And definitely add a test for what's supposed to happen in this case so it's clear to a reader what should/does happen if you try to send over a workload with a space in it. The easiest thing might be to send as "{replyQueue} {index} {experiment}" and then parse with SplitN(reply, 3).

workloads/rest_test.go
```diff
@@ -121,9 +121,9 @@ var _ = Describe("Rest Workloads", func() {
 				client.ShouldHaveBeenCalledWith("POST(uaa)", "THELOGINSERVER/PATH/oauth/token")
 			})
 
-			Context("When a username and password are configured", func() {
+			Context("When username and password are configured", func() {
```
julz: Just for completeness, how about if we do this as an additional test (when multiple usernames and passwords are configured) and keep the existing test for when just one user/pass are configured (I know that's actually a special case of this in a sense, but I think it's clearer to test the simple case and then the more complex one).

workloads/rest_test.go
```diff
@@ -141,6 +141,28 @@ var _ = Describe("Rest Workloads", func() {
...
+					Ω(data.(url.Values)["password"]).Should(Equal([]string{"pass1"}))
+
+					context["workerIndex"] = 2
+					rest.Login(context)
+					data = client.ShouldHaveBeenCalledWith("POST(uaa)", "THELOGINSERVER/PATH/oauth/token")
+					Ω(data.(url.Values)["username"]).Should(Equal([]string{"user2"}))
+					Ω(data.(url.Values)["password"]).Should(Equal([]string{"pass2"}))
+				})
+
+				It("recycles the list of username and password when there are more workers than username", func() {
```
julz: can we add a quick test for what happens when there are a different number of passwords from usernames? seems like a common case to have a few test users all with the same password, be nice to explicitly state that that works in the tests.

julz:
after saying gtg I thought of an additional couple tests it does seem worth adding, which completely makes me a liar, sorry :) Anyway, the main thing is the comment about being sure to handle spaces in the experiment string because the redis module shouldn't assume things about what strings the worker module may or may not accept; if you also could add the extra couple tests above while you're there that'd be great tho :)

simonleung8:
Updated the tests in rest_test.go
also fixed redis.go so spaces in experiment string will not cause error

bravera:
Partially works although pushed apps never seem to start properly (they remain shown in red text if you run 'cf apps').
If you include invalid credentials as first in the list then errors are correctly thrown. If you put invalid credentials later in the list then no errors are thrown and any pushes are done

PR #75, opened 2014-04-15
Go steno!

julz:
Convert all the non-UI fmt.Printlns to steno, because: steno! 

(also because it should make debugging with the animated console UI on much easier, because you can divert logs to a file now with -logging:file=foo.log)


simonleung8:
Everything looks go, rebased and merged

PR #78, opened 2014-04-22
allow rest:push to be used in Web UI

simonleung8:
Web UI is now possible to specify a list of operations to run a custom experiment and it also prevents user to create an invalid experiment.
- I don't quite like the arrangement of the configuration box now, what we have will do for now, but I will re-arrange the inputs at a later time so it will be more pleasant to look at

### DIFF
--- a/benchmarker/benchmarker.go
+++ b/benchmarker/benchmarker.go
@@ -3,6 +3,8 @@ package benchmarker
 import (
 	"sync"
 	"time"
+
+	"github.com/cloudfoundry-incubator/pat/context"
 )
 
 type StepResult struct {
@@ -17,36 +19,36 @@ type IterationResult struct {
 }
 
 func Time(experiment func() error) (result time.Duration, err error) {
-	t0 := time.Now()
+	t0 := time.Now()	
 	err = experiment()
 	t1 := time.Now()
 	return t1.Sub(t0), err
 }
 
-func Counted(out chan<- int, fn func(map[string]interface{})) func(map[string]interface{}) {
-	return func(workloadCtx map[string]interface{}) {
+func Counted(out chan<- int, fn func(context.WorkloadContext)) func(context.WorkloadContext) {
+	return func(workloadCtx context.WorkloadContext) {
 		out <- 1
 		fn(workloadCtx)
 		out <- -1
 	}
 }
 
-func TimedWithWorker(out chan<- IterationResult, worker Worker, experiment string) func(map[string]interface{}) {
-	return func(workloadCtx map[string]interface{}) {
+func TimedWithWorker(out chan<- IterationResult, worker Worker, experiment string) func(context.WorkloadContext) {
+	return func(workloadCtx context.WorkloadContext) {
 		time := worker.Time(experiment, workloadCtx)
 		out <- time
 	}
 }
 
-func Once(fn func(map[string]interface{})) <-chan func(map[string]interface{}) {
+func Once(fn func(context.WorkloadContext)) <-chan func(context.WorkloadContext) {
 	return Repeat(1, fn)
 }
 
-func RepeatEveryUntil(repeatInterval int, runTime int, fn func(map[string]interface{}), quit <-chan bool) <-chan func(map[string]interface{}) {
+func RepeatEveryUntil(repeatInterval int, runTime int, fn func(context.WorkloadContext), quit <-chan bool) <-chan func(context.WorkloadContext) {
 	if repeatInterval == 0 || runTime == 0 {
 		return Once(fn)
 	} else {
-		ch := make(chan func(map[string]interface{}))
+		ch := make(chan func(context.WorkloadContext))
 		var tickerQuit *time.Ticker
 		ticker := time.NewTicker(time.Duration(repeatInterval) * time.Second)
 		if runTime > 0 {
@@ -72,8 +74,8 @@ func RepeatEveryUntil(repeatInterval int, runTime int, fn func(map[string]interf
 	}
 }
 
-func Repeat(n int, fn func(map[string]interface{})) <-chan func(map[string]interface{}) {
-	ch := make(chan func(map[string]interface{}))
+func Repeat(n int, fn func(context.WorkloadContext)) <-chan func(context.WorkloadContext) {
+	ch := make(chan func(context.WorkloadContext))
 	go func() {
 		defer close(ch)
 		for i := 0; i < n; i++ {
@@ -83,31 +85,23 @@ func Repeat(n int, fn func(map[string]interface{})) <-chan func(map[string]inter
 	return ch
 }
 
-func Execute(tasks <-chan func(map[string]interface{}), workloadCtx map[string]interface{}) {
+func Execute(tasks <-chan func(context.WorkloadContext), workloadCtx context.WorkloadContext) {
 	for task := range tasks {		
 		task(workloadCtx)
 	}
 }
 
-func ExecuteConcurrently(workers int, tasks <-chan func(map[string]interface{}), workloadCtx map[string]interface{}) {
+func ExecuteConcurrently(workers int, tasks <-chan func(context.WorkloadContext), workloadCtx context.WorkloadContext) {
 	var wg sync.WaitGroup
 	for i := 0; i < workers; i++ {
 		wg.Add(1)
-		workloadCtx["workerIndex"] = i
-		go func(t <-chan func(map[string]interface{}), ctx map[string]interface{}) {
+		workloadCtx.PutInt("workerIndex", i)
+		go func(t <-chan func(context.WorkloadContext), ctx context.WorkloadContext) {
 			defer wg.Done()
 			for task := range t {
 				task(ctx)
 			}
-		}(tasks, clone(workloadCtx))
+		}(tasks, workloadCtx.Clone())
 	}
 	wg.Wait()
-}
-
-func clone(src map[string]interface{}) map[string]interface{} {
-	var clone = make(map[string]interface{})
-	for k, v := range src {
-    	clone[k] = v
-	}
-	return clone
 }
--- a/benchmarker/benchmarker.go
+++ b/benchmarker/benchmarker.go
@@ -23,30 +23,30 @@ func Time(experiment func() error) (result time.Duration, err error) {
 	return t1.Sub(t0), err
 }
 
-func Counted(out chan<- int, fn func(int)) func(int) {
-	return func(workerIndex int) {
+func Counted(out chan<- int, fn func(map[string]interface{})) func(map[string]interface{}) {
+	return func(workloadCtx map[string]interface{}) {
 		out <- 1
-		fn(workerIndex)
+		fn(workloadCtx)
 		out <- -1
 	}
 }
 
-func TimedWithWorker(out chan<- IterationResult, worker Worker, experiment string) func(int) {
-	return func(workerIndex int) {
-		time := worker.Time(experiment, workerIndex)
+func TimedWithWorker(out chan<- IterationResult, worker Worker, experiment string) func(map[string]interface{}) {
+	return func(workloadCtx map[string]interface{}) {
+		time := worker.Time(experiment, workloadCtx)
 		out <- time
 	}
 }
 
-func Once(fn func(int)) <-chan func(int) {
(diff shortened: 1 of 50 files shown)


benchmarker/redis.go
```diff
@@ -78,16 +97,21 @@ func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
...
 		if err == nil {
-			parts := strings.SplitN(reply[1], " ", 3)
-
-			go func(experiment string, replyTo string, workerIndex string) {
-				i, _ := strconv.Atoi(workerIndex)
-				result := delegate.Time(experiment, i)
+			parts := strings.SplitN(reply[1], delimiter, 6)
+			workloadCtx := make(map[string]interface{})
+			workerIndex, _ := strconv.Atoi(parts[1])
+			workloadCtx["workerIndex"] = workerIndex
```
julz: hmmm I don't think redis should be whitelisting specific keys like this.. couples those two modules together a bit tightly..

benchmarker/benchmarker.go
```diff
@@ -83,22 +83,31 @@ func Repeat(n int, fn func(int)) <-chan func(int) {
...
-func Execute(tasks <-chan func(int)) {
+func Execute(tasks <-chan func(map[string]interface{}), workloadCtx map[string]interface{}) {
 	for task := range tasks {
-		task(1)
+		task(workloadCtx)
 	}
 }
 
-func ExecuteConcurrently(workers int, tasks <-chan func(int)) {
+func ExecuteConcurrently(workers int, tasks <-chan func(map[string]interface{}), workloadCtx map[string]interface{}) {
```
julz: unless I missed it, there's no test that shows that if you pass a context map to the function it is made available to the tasks (should be one for local and one for redis showing that works)

benchmarker/redis.go
```diff
@@ -27,9 +27,26 @@ func NewRedisWorkerWithTimeout(conn redis.Conn, timeoutInSeconds int) Worker {
...
 	guid, _ := uuid.NewV4()
-	rw.conn.Do("RPUSH", "tasks", "replies-"+guid.String()+" "+strconv.Itoa(workerIndex)+" "+experiment)	
+
+	if (workloadCtx["cfTarget"] == nil) { workloadCtx["cfTarget"] = "" }
+	if (workloadCtx["cfUsername"] == nil) { workloadCtx["cfUsername"] = "" }
+	if (workloadCtx["cfPassword"] == nil) { workloadCtx["cfPassword"] = "" }
+	if (workloadCtx["workerIndex"] == nil) { workloadCtx["workerIndex"] = 0 }
+	workerIndex := strconv.Itoa(workloadCtx["workerIndex"].(int))
+	
+	//there should not be any spaces in the strings, but just in case, using a special delimiter
```
julz: how does this help? it just makes it harder to hit the bug (because the string now needs to contain '% %'). It would be better to escape the strings at the other side, and blow up on this side. Also, unless I missed it, I don't see a test for this.

simonleung8:
Code updated:
- Test for both Execute() and ExecuteConcurrently(), making sure context map is passed to the test function
- Other than the input validation in UI, server.go now removes spaces in the argument inputs.
- redis.go returns error if either cfTarget, cfUsername, cfPassword and workload contain spaces.
- redis.go no longer use a special delimiter to separate arguments
- cmdline.go removes spaces in -workload paramenter

benchmarker/benchmarker.go
simonleung8: now there is test for this in benchmarker.go

benchmarker/redis.go
```diff
@@ -27,9 +28,28 @@ func NewRedisWorkerWithTimeout(conn redis.Conn, timeoutInSeconds int) Worker {
...
+	if (workloadCtx["cfPassword"] == nil) { workloadCtx["cfPassword"] = "" }
+	if (workloadCtx["workerIndex"] == nil) { workloadCtx["workerIndex"] = 0 }
+	workerIndex := strconv.Itoa(workloadCtx["workerIndex"].(int))
+
+	if (strings.Contains(workloadCtx["cfTarget"].(string)," ") || strings.Contains(workloadCtx["cfUsername"].(string)," ") || 
+		strings.Contains(workloadCtx["cfPassword"].(string)," ") || strings.Contains(experiment," ")) {
+		return IterationResult{0, []StepResult{}, encodeError(errors.New("Redis worker error: workload, cfTarget, cfUsername or cfPassword cannot contain space in the string"))}
+	}
+	
+	redisStr := "replies-"+guid.String() +
```
julz: I really don't think redis.go should have any idea about what keys might be in the context map. This is going to cause hard to debug problems when keys randomly disappear in redis mode as soon as someone forgets that redis silently strips any keys not hardcoded here (we're very likely to add new keys over time).  Let's find a way to do this that doesn't involve redis.go knowing what keys might be inside the context map?

benchmarker/redis_test.go
```diff
@@ -131,6 +133,41 @@ var _ = Describe("RedisWorker", func() {
...
+
+	Describe("Rest APIs argument in workloadCtx and workload string contain spaces", func() {
+		AfterEach(func() {
+			workloadCtx["cfTarget"] = ""
+			workloadCtx["cfUsername"] = ""
+			workloadCtx["cfPassword"] = ""			
+		})
+
+		It("return errors when workload string contains space", func() {
+			worker := NewRedisWorker(conn)
```
julz: when I said we should escape any problematic strings before they got to the redis slave, I meant in the redis stub (i.e. before pushing onto the redis queue we should escape any spaces). Having server.go explicitly check for spaces just because redis.go can't handle them makes no sense; what about cmdline.go? what about any other UIs we might add in the future that might use redis? 

I have no problem with server.go doing some validation and returning some errors and so forth, but this is the wrong layer to handle redis-related escaping. 

(also as a very very nit-picky thing, It "returns errors", not It "return errors" :)

cmdline/cmdline.go
```diff
@@ -34,9 +35,11 @@ func InitCommandLineFlags(config config.Config) {
 	config.BoolVar(&params.listWorkloads, "list-workloads", false, "Lists the available workloads")
 	benchmarker.DescribeParameters(config)
 	store.DescribeParameters(config)
+	// flags not set here
```
julz: what is this comment trying to achieve, because what it achieved was confusing me

cmdline/cmdline_test.go
```diff
@@ -22,7 +22,7 @@ var _ = Describe("Cmdline", func() {
 
 	BeforeEach(func() {
 		WithConfiguredWorkerAndSlaves = func(fn func(Worker benchmarker.Worker) error) error {
-			worker := benchmarker.NewLocalWorker()
+			worker := benchmarker.NewLocalWorker()
```
julz: go fmt (in a few places)

server/server.go
```diff
@@ -124,15 +125,25 @@ func (ctx *context) handlePush(w http.ResponseWriter, r *http.Request) (interfac
...
-		workload = "push"
+		workload = "gcf:push"
 	}
 
+	workloadContext := make(map[string]interface{})
+	workloadContext["cfTarget"] = cfTarget
+	workloadContext["cfUsername"] = cfUsername
+	workloadContext["cfPassword"] = cfPassword
+
+	//init context here and eventually context will be sent to ExecuteConcurrently()
```
julz: why do you care in server.go how this eventually gets run? if the line below is unclear, we should try to make it clearer; I don't think this comment particularly does that

server/server.go
```diff
@@ -124,15 +125,25 @@ func (ctx *context) handlePush(w http.ResponseWriter, r *http.Request) (interfac
 		stop = 0
 	}
 
-	workload := r.FormValue("workload")
+	cfTarget := removeSpaces( r.FormValue("cfTarget") )
+	cfUsername := removeSpaces( r.FormValue("cfUsername") )
+	cfPassword := removeSpaces( r.FormValue("cfPassword") )
```
julz: what if the password actually does have a space?

benchmarker/redis.go
simonleung8: We could loop through the map, escape all the strings and push everything into redis, but what if in the future there are fields in the context map we do not wish to push into redis?

simonleung8: another way of doing it is passing an extra field in the context map containing an array of keys to be pushed into redis.  With this approach, we still need to have this array of keys defined somewhere

julz: when would we ever want something in the context map to silently disappear when used with redis?

simonleung8:
Code updated
- Redis now allows the content in context maps to contain spaces.
- As for now, Redis passes over all the context that is listed in benchmaker.config.go, we currently only pass over 3 keys (username, password and target).  This approach seems to be suffice for now.
  When we need to deal with more keys, we certainly could loop through the entire context map, pass over all the keys also the types with redis.

server/server.go
julz: @simonleung8  - did you see the above comment?

benchmarker/redis.go
```diff
@@ -19,6 +19,8 @@ type rw struct {
 
 const DEFAULT_TIMEOUT = 60 * 5
```
julz: oops, this is my bad, too much Java in my youth ;) this should be mixed case, like spaceEscapeStr below, do you mind updating that while you have the file open?

benchmarker/redis.go
```diff
@@ -95,3 +113,53 @@ func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
 		}
 	}
 }
+
+func replaceSpaceWithEscape(workloadCtx map[string]interface{}, escapeStr string) map[string]interface{} {
```
julz: rather than reinventing this, let's use something from the standard library like [url.QueryEscape](http://golang.org/pkg/net/url/#QueryEscape)

benchmarker/config.go
```diff
@@ -8,6 +8,8 @@ import (
 	"github.com/cloudfoundry-community/pat/workloads"
 )
 
+var RedisContextMapStr = []string{"cfTarget", "cfUsername", "cfPassword"}
```
julz, line 11: I still have no idea what this buys us. When does this do something other than causing a silent, _really_ hard to find bug? When do we ever want something to silently disappear from the context map before being sent to the redis workers? Could we add this Acme-5000-Super-Effective-Silent-Bug-Creator module when we actually have a field in the context that we _don't_ want to get passed to redis?

simonleung8:
Updated redis.go to accept everything in the context map.  Also spaces are allowed in the content.

simonleung8:
not sure why github can't auto merge this PR, I tried to rebase with master and it wanted to rebase this with the some old branch (mine is 10 behind?).  I guess it will be easier to re-upload this branch after you done reviewing so we don't lose the comments now.

benchmarker/redis.go
```diff
@@ -95,3 +102,68 @@ func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
 		}
 	}
 }
+
+func turnContextToRedisStr(workloadCtx map[string]interface{})(strValue string, strType string, strKey string){
```
julz: why can't we just send over as json? e.g. create a `struct redisMessage { reply string; workload string; context map[string]interface{} }`, whack the reply, workload and context into the struct, and then have the stub encode and the slave decode..

server/server.go
```diff
@@ -124,15 +125,24 @@ func (ctx *context) handlePush(w http.ResponseWriter, r *http.Request) (interfac
 		stop = 0
 	}
 
-	workload := r.FormValue("workload")
+	cfTarget := removeSpaces( r.FormValue("cfTarget") )
+	cfUsername := removeSpaces( r.FormValue("cfUsername") )
```
julz: why do we silently strip spaces from usernames? shouldn't we error instead? ditto if the target has a space.. this is an API with a guy in front, in what circumstance are strings with spaces happening, and why isn't it just an error?

simonleung8: Updated.
on a second thought, I don't want the server to remove any spaces since the UI is already checking for invalid entry, which is much more user friendly.
Though I think it is appropriate to trim leading and trailing spaces from the input (UI should also prevent leading and trailing spaces).

simonleung8:
updated with a package context to handle workload context throughput steps

simonleung8:
Superseded by PR #91

PR #79, opened 2014-04-22
Experiment Configuration Popup

simonleung8:
A popup for experiment configuration

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -41,9 +41,9 @@ body { padding: 8px; }
     <div class="panel-heading">
       <button class="btn btn-primary btn-xs" style="float: right" data-toggle="modal" data-target="#historyPopup">
         <span class="glyphicon glyphicon-time"></span> Histories
-      </button>      
-      <button class="btn btn-primary btn-xs" data-toggle="modal" data-target="#experimentPopup">        
-        <span class="glyphicon glyphicon-flash"></span> Run Experiment
+      </button>
+      <button class="btn btn-primary btn-xs" data-toggle="modal" data-target="#experimentPopup">
+        <span class="glyphicon glyphicon-flash"></span> Run Experiment..
       </button>
     </div>
     <div id="graph" class="panel-body col-md-12 center-block" data-bind="chart: data" style="">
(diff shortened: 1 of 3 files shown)


ui/index.html
```diff
@@ -43,6 +43,10 @@
       <button class="btn btn-primary btn-xs" style="float: right" data-toggle="modal" data-target="#historyPopup">
         <span class="glyphicon glyphicon-time"></span> Histories
       </button>
+      <span style="float: right">&nbsp&nbsp|&nbsp&nbsp</span>
+      <button class="btn btn-primary btn-xs" style="float: right" data-toggle="modal" data-target="#experimentPopup">        
+        <span class="glyphicon glyphicon-flash"></span> Configure Experiment
```
julz: surely "Run Experiment" is better? Also the button is kind of hard to find on the right given it's the thing 100% of users want to do first. And after hitting "start" the panel doesn't disappear which is annoying.

ui/index.html
```diff
@@ -83,56 +87,61 @@
...
-          <div class="col-sm-6">
-            <input type="number" class="form-control" id="inputInterval" name="inputInterval" placeholder="0" data-bind="value: numInterval">
-          </div>
-        </div>
-        <div class="form-group" data-bind="css: { 'has-error': numStopHasError }">
-          <label for="inputStop" class="col-sm-2 control-label">Stop</label>
-          <div class="col-sm-6">
-            <input type="number" class="form-control" id="inputStop" name="inputStop" placeholder="0" data-bind="value: numStop">
-          </div>
+  <div class="modal fade" id="experimentPopup" tabindex="-1" role="dialog" aria-labelledby="experimentPopupLabel" aria-hidden="true" >
```
julz: still not a fan of the fading, I don't think it's helpful

simonleung8:
Code updated.  
Button is now at the left, easier to spot
"Start" button will also dismiss the model form
No more fade, just a plain popup now

julz:
looks good, I added a quick commit to change the text to "Run Experiment .." to make it clear that you'll get some options before being able to run it. also got rid of the download csv button (which you can still get to from the history list) and stop experiment button (which never worked anyway), which no longer make sense with a modal experiment run dialog.

PR #83, opened 2014-04-25
improve screen estate for content

simonleung8:
Removed the large "PAT" title and made the control buttons a little bigger.

This allows more screen space for contents

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -28,24 +28,17 @@ body { padding: 8px; }
 </head>
 
 <body>
-
 <div class="container">
-  <div class="header row">
-    <h1>Performance Acceptance Tests <small>v0.1</small></h1>
-  </div>
-
-  <div class="row">
-  </div>
-
   <div class="row panel panel-default">
     <div class="panel-heading">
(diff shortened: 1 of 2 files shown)


ui/index.html
```diff
@@ -28,24 +28,17 @@
...
-  </div>
-
-  <div class="row">
-  </div>
-
   <div class="row panel panel-default">
     <div class="panel-heading">
-      <button class="btn btn-primary btn-xs" style="float: right" data-toggle="modal" data-target="#historyPopup">
-        <span class="glyphicon glyphicon-time"></span> Histories
+      <font class=" patTitle">PERFORMANCE ACCEPTANCE TEST</font>
```
julz: Font isn't really a tag we should use in 2014 (use something like h2 ideally, or failing that span). Also would be better to have this be regular case, and then use text-transform: uppercase in the css. Finally, should be "Performance Acceptance Tests" (plural)

julz:
looks good, though we could probably do better on the reducing wasted space front eventually; made a comment in the code that would be good to fix up before merging

simonleung8:
Updated,  <font> is replaced with h3, 
and added text-transform

PR #84, opened 2014-04-26
Correct interface for Gomega Match

simonleung8:
ginkgo in /cmdline is not running because we incorrectly implemented Gomega Match method 
Current Code: Match(actual interface{}) (success bool, err error)

The correct interface for Gomega Match is 
Match(actual interface{}) (success bool, message string, err error)

*source: http://onsi.github.io/gomega/#adding_your_own_matchers


simonleung8:
looks like my ginkgo/gomega version is different than the one travis downloads

PR #85, opened 2014-04-29
Enable gcf:GenerateAndPush in Web UI

simonleung8:
Enabling gcf:GenerateAndPush in Web UI


simonleung8:
This is not needed anymore

PR #87, opened 2014-04-30
Added test cases to demo rest HTTP calls

ghost:
I split the test cases into separate blocks, which makes it a little hard to read on github. #L278 is where my test cases actually start.


julz:
looks great, for future reference if you want to make the diff clear when you've done a refactoring and then added stuff, you can just split it into two commits, do the refactor in the first and then the adds in the second

PR #69, opened 2014-04-08
Allow user to ramp workers up/down each iteration

jberkhahn:
modified ExecuteConcurrently() to limit workers throughput based on a supplied pattern in the form of a comma separated list, i.e. -pattern=1,2,3. Default behavior in the case of a missing/malformed pattern is to simply run each iteration with all the workers.

PR #89, opened 2014-05-07
Added ramping example to ReadMe. Added error-checking for invalid input for -concurrency

jberkhahn:
Quick touch up for PR #69. Added an example of how to use the ramping functionality to the readme. Also, added error checking for -concurrency as it is now parsed as in as a string, and thus errors won't be caught by the flag parsing library.


julz:
looks good, but you typoed `timeBetweenStep`, should be `timeBetweenSteps` (plural)

jberkhahn:
Fixed.

PR #90, opened 2014-05-08
Write experiment meta data to file and redis

ghost:
Match the guid for each experiments csv/redis file to the meta data file.


julz:
breaks travis

benchmarker/redis_test.go
```diff
@@ -3,11 +3,9 @@ package benchmarker
 import (
 	"errors"
 	"io"
-	"os/exec"
-	"path/filepath"
-	"runtime"
 	"time"
 
+	"github.com/cloudfoundry-incubator/pat/ginkgo/redis_helpers"
```
julz: what does this have to do with ginkgo? maybe a 'test-helpers/redis' package?

cmdline/cmdline.go
```diff
@@ -37,6 +38,7 @@ func InitCommandLineFlags(config config.Config) {
 	config.IntVar(&params.interval, "interval", 0, "repeat a workload at n second interval, to be used with -stop")
 	config.IntVar(&params.stop, "stop", 0, "stop a repeating interval after n second, to be used with -interval")
 	config.BoolVar(&params.listWorkloads, "list-workloads", false, "Lists the available workloads")
+	config.StringVar(&params.note, "note", "", "Description about the experiment")
```
julz: if it's a description, why isn't the flag called "description"?

cmdline/cmdline.go
```diff
@@ -44,12 +46,12 @@ func InitCommandLineFlags(config config.Config) {
 func RunCommandLine() error {
 	return WithConfiguredWorkerAndSlaves(func(worker benchmarker.Worker) error {
 		return validateParameters(worker, func() error {
-			return store.WithStore(func(store Store) error {
+			return store.WithStore(func(lab_store Store) error {
```
julz: why?

cmdline/cmdline.go
```diff
@@ -58,11 +60,16 @@ func RunCommandLine() error {
 					})
 				}
 
-				lab.RunWithHandlers(
+				guid, _ := lab.RunWithHandlers(
 					NewRunnableExperiment(
 						NewExperimentConfiguration(
 							params.iterations, parsedConcurrency, parsedConcurrencyStepTime, params.interval, params.stop, worker, params.workload)), handlers)
 
+				meta, _ := store.MetaStoreFactory("output/meta")
```
julz: can we avoid this duplication? (you need to create a MetaStore and add data to it here and in server). Why not just add a description string to `lab.RunWithHandlers` and move the function in there?

ginkgo/redis_helpers/redis.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"os/exec"
+	"path"
+	"path/filepath"
+	"runtime"
+	"time"
+)
+
+func StartRedis(config string) {
```
julz: the package is called 'redis_helpers', so we don't need the word 'Redis' in all these methods

server/server.go
```diff
@@ -142,6 +145,10 @@ func (ctx *context) handlePush(w http.ResponseWriter, r *http.Request) (interfac
 			NewExperimentConfiguration(
 				pushes, concurrency, concurrencyStepTime, interval, stop, ctx.worker, workload)))
 
+	if ctx.meta != nil {
```
julz: why is it ever nil? do we want to support not storing metadata? isn't it better to just have the store have an extra method?

store/config.go
```diff
@@ -44,3 +44,7 @@ var RedisStoreFactory = func(conn redis.Conn) (laboratory.Store, error) {
 var CsvStoreFactory = func(dir string) laboratory.Store {
 	return NewCsvStore(dir, workloads.DefaultWorkloadList())
 }
+
+func MetaStoreFactory(directory string) (*MetaStore, error) {
```
julz: why is the "metastore" different from the actual store? is it ever a good idea to store metadata in a different store from the experiment data? can't the store just implement both interfaces (or can't the store interface just  have an extra method?)

store/config_test.go
```diff
@@ -9,75 +10,136 @@ import (
...
+	Describe("#MetaStoreFactory", func() {
+		const (
+			dir = "tmp"
+		)
+
+		var (
+			meta *MetaStore
+		)
+
+		It("Returns a pointer to a MetaStore struct", func() {
```
julz: yuck, this is pretty low-level for a test

store/meta.go
```diff
@@ -0,0 +1,107 @@
...
+		strconv.Itoa(meta_data.Interval), meta_data.Workload, meta_data.Note}
+
+	writer.Write(header)
+	writer.Write(body)
+	writer.Flush()
+
+	return nil
+}
+
+func (self *MetaStore) writeRedis(name string, meta_data MetaData) error {
```
julz: ouch. no. let's have each of the existing stores implement a `writeMetadata` method (or something). that way we don't have to add a new method to this class whenever we add a store

julz:
ideally would be good to have the refactoring (moving the StartRedis stuff to a helper package) be in a separate commit from the new function; don't worry tons about this, but it'd be nice to have a clear commit history of small commits that do one thing

redis/conn_test.go
```diff
@@ -1,10 +1,8 @@
 package redis_test
 
 import (
-	"os/exec"
-	"path/filepath"
-	"runtime"
-	"time"
+	redisHelpers "github.com/cloudfoundry-incubator/pat/test_helpers/redis"
```
julz: this is bad style, would be better to rename the package if it's going to clash with things... how about we just have a 'helpers' package with redis.go a file inside that?

server/server_test.go
```diff
@@ -82,7 +86,7 @@ var _ = Describe("Server", func() {
 		csvs := store.NewCsvStore("/var/tmp/foo/bar/", workloads.DefaultWorkloadList())
 		ch := make(chan *Sample)
 		go func() { ch <- &Sample{}; ch <- &Sample{}; close(ch) }()
-		csvs.Writer("1234")(ch)
+		csvs.Writer("1234", ExperimentConfiguration{})(ch)
```
julz: why have a guid and a config as separate object? seems better to just pass an object that represents the experiment and has the guid

store/config.go
```diff
@@ -44,3 +44,7 @@ var RedisStoreFactory = func(conn redis.Conn) (laboratory.Store, error) {
 var CsvStoreFactory = func(dir string) laboratory.Store {
 	return NewCsvStore(dir, workloads.DefaultWorkloadList())
 }
+
+/*func MetaStoreFactory(directory string) (*MetaStore, error) {
```
julz: make sure we lose the dead code

.travis.yml
```diff
@@ -16,7 +16,7 @@ before_script:
   # - cd casperjs; git checkout -q tags/1.1-beta1; cd ../
   # - export PATH=$PATH:`pwd`/casperjs/bin
   - gocart
-  - go build github.com/onsi/ginkgo/ginkgo
+  #- go build github.com/onsi/ginkgo/ginkgo
```
julz: does something break without this commented? either we should remove it, or leave it in, but let's not leave it commented without explanation

experiment/runner.go
```diff
@@ -107,6 +108,10 @@ func newRunningExperiment(iterations int, iterationResults chan IterationResult,
 	return &SamplableExperiment{iterations, iterationResults, workers, samples, quit}
 }
 
+func (config *RunnableExperiment) GetExperimentConfiguration() ExperimentConfiguration {
```
julz: is there a way we could "tell don't ask" here.. e.g could we ask the experiment to describe itself, e.g. add a `Describe(metadata Metadata)` method rather than pulling back the experimentconfiguration object?

store/csv.go
```diff
@@ -48,7 +50,61 @@ func (file *csvFile) AddWorkloadStep(workload workloads.WorkloadStep) {
...
+func (self *CsvStore) writeMeta(startTime time.Time, guid string, ex experiment.ExperimentConfiguration) {
+	var logger = logs.NewLogger("store.meta")
+
+	dir := path.Join(self.dir, "csv.meta")
+
+	var writer *csv.Writer
+
+	file, err := os.OpenFile(dir, os.O_RDWR, 0755)
+	if os.IsNotExist(err) {
+		logger.Infof("Creating directory, %s", filepath.Dir(dir))
```
julz: let's pull this out in to a method e.g. OpenOrCreate

store/csv.go
```diff
@@ -48,7 +50,61 @@ func (file *csvFile) AddWorkloadStep(workload workloads.WorkloadStep) {
...
+
+	//only write the header once
+	if len(lines) == 0 {
+		header := []string{"csv guid", "start time", "iterations", "concurrency",
+			"concurrency step time", "stop", "interval", "workload", "note"}
+		writer.Write(header)
+	}
+
+	var concurrency string
+	for iter, value := range ex.Concurrency {
```
julz: why not just [strings.Join](http://golang.org/pkg/strings/#Join)?

experiment/runner.go
julz: ^^ did you miss this comment? I'd rather we didn't have a GetExperimentConfiguration() on RunnableExperiment, because then there's really no point having RunnableExperiment be separate from ExperimentConfiguration.. how about a Describe(..) method on RunnableExperiment instead?

julz: (or even just have GetExperimentConfiguration return a map, but I'm in general not a huge fan of getters, I think it's nicer to ask an object to do something rather than grabbing its internal state)

store/csv.go
```diff
@@ -185,12 +226,14 @@ func (store *CsvStore) LoadAll() (samples []experiment.Experiment, err error) {
 
 	samples = make([]experiment.Experiment, 0)
 	for _, f := range files {
-		base := strings.Split(f.Name(), ".")[0]
-		name := strings.SplitN(base, "-", 2)[1]
-		if len(name) > 0 {
-			loaded, err := store.load(f.Name(), name)
-			if err == nil {
-				samples = append(samples, loaded)
+		if filepath.Ext(f.Name()) == ".csv" {
```
julz, line 220: so if the file doesn't end in .csv it silently fails? is this what we want and is there a test I missed for this somewhere?

ghost, line 220: This is here to simply exclude the .meta data files when reading in the data. Right now the .csv files and .meta data file are written to the same directory.

experiment/runner.go
ghost: The problem is in Laboratory.go file, that is where we need the ExperimentConfiguration{}. This way we can pass the data to the csv/redis writer. However, we are returning this object through an interface, and so that interface needs a function to obtain the actual data, not simply a run function on the struct itself where it has access to the data.

store/csv.go
julz, line 220: is there a test for it? (I might've missed it)

experiment/runner.go
julz: but why does it need to return the ExperimentConfiguration object, rather than having the method either a) convert it to a map instead (`func GetMetadata() map[string]string`)  or b) even better, having a `Describe(..)` method that can tell a Metadata interface about the fields it wants to store.. the method that implements that will be in laboratory.go, and will have access to the ExperimentConfiguration, but we won't be exposing the ExperimentConfiguration out everywhere

experiment/runner.go
```diff
@@ -107,6 +113,30 @@ func newRunningExperiment(iterations int, iterationResults chan IterationResult,
...
+	mMap["guid"] = e.Guid
+	mMap["description"] = e.Description
+	mMap["iterations"] = strconv.Itoa(e.Iterations)
+	mMap["concurrency step time"] = e.ConcurrencyStepTime.String()
+	mMap["interval"] = strconv.Itoa(e.Interval)
+	mMap["stop"] = strconv.Itoa(e.Stop)
+	mMap["workload"] = e.Workload
+
+	var concurrency string
+	for iter, value := range e.Concurrency {
```
julz: lets just make this concurrency.start and concurrency.end rather than converting in to a "4..5" string; I'd like to eventually have "4..5" just be shorthand for concurrent.start=4, concurrency.end=5

laboratory/laboratory_test.go
```diff
@@ -175,3 +175,10 @@ func (e *dummyExperiment) GetData() ([]*Sample, error) {
 func (e *dummyExperiment) GetGuid() string {
 	return e.name
 }
+
+func (e *dummyExperiment) DescribeMetadata() map[string]string {
+	mMap := NewExperimentConfiguration(0, nil, 0, 0, 0, nil, "", "").DescribeMetadata()
+	//make(map[string]string)
+	//mMap["guid"] = "guid-1"
```
julz: ?

store/csv.go
```diff
@@ -48,20 +51,49 @@ func (file *csvFile) AddWorkloadStep(workload workloads.WorkloadStep) {
...
+	}
+
+	//only write the header once
+	if len(lines) == 0 {
+		header := []string{"csv guid", "start time", "iterations", "concurrency",
+			"concurrency step time", "stop", "interval", "workload", "description"}
+		writer.Write(header)
+	}
+
+	body := []string{meta["guid"], startTime.Format(time.RFC850), meta["iterations"],
```
julz, line 83: why don't we just json/yml encode the map rather than relying on knowing what fields it contains? that way DescribeMetadata can return whatever information it likes and we don't have to update both places

.travis.yml
```diff
@@ -16,7 +16,6 @@ before_script:
   # - cd casperjs; git checkout -q tags/1.1-beta1; cd ../
   # - export PATH=$PATH:`pwd`/casperjs/bin
   - gocart
-  - go build github.com/onsi/ginkgo/ginkgo
```
julz, line 19: what's the deal with this?

ghost, line 19: This line had caused ginkgo to fail before. Not sure why that was exactly happening, seems to be fixed now.

PR #91, opened 2014-05-08
package context & rest:api for web ui

simonleung8:
- A type safe context map package, used for sharing information between worker steps
- web ui can now use rest:api

### DIFF
--- a/benchmarker/redis.go
+++ b/benchmarker/redis.go
@@ -2,9 +2,8 @@ package benchmarker
 
 import (
 	"encoding/json"
-	"strings"
-	"strconv"
 
+	"github.com/cloudfoundry-incubator/pat/context"
 	"github.com/cloudfoundry-incubator/pat/logs"
 	"github.com/cloudfoundry-incubator/pat/redis"
 	"github.com/cloudfoundry-incubator/pat/workloads"
@@ -17,19 +16,40 @@ type rw struct {
 	timeoutInSeconds int
 }
 
-const DEFAULT_TIMEOUT = 60 * 5
+type redisMessage struct {
+	Reply           string
+	Workload        string
+	WorkloadContext context.Context
+}
+
+const DefaultTimeout = 60 * 5
 
 func NewRedisWorker(conn redis.Conn) Worker {
-	return NewRedisWorkerWithTimeout(conn, DEFAULT_TIMEOUT)
+	return NewRedisWorkerWithTimeout(conn, DefaultTimeout)
 }
 
 func NewRedisWorkerWithTimeout(conn redis.Conn, timeoutInSeconds int) Worker {
 	return &rw{defaultWorker{make(map[string]workloads.WorkloadStep)}, conn, timeoutInSeconds}
 }
 
-func (rw rw) Time(experiment string, workerIndex int) (result IterationResult) {
+func (rw rw) Time(workload string, workloadCtx context.Context) (result IterationResult) {
 	guid, _ := uuid.NewV4()
-	rw.conn.Do("RPUSH", "tasks", "replies-"+guid.String()+" "+strconv.Itoa(workerIndex)+" "+experiment)	
+	redisMsg := redisMessage{
+		Workload:        workload,
+		Reply:           "replies-" + guid.String(),
+		WorkloadContext: workloadCtx,
+	}
+	
+	var jsonRedisMsg []byte
+	var err error
+	jsonRedisMsg, err = json.Marshal(redisMsg)
+
+	if err != nil {
+		return IterationResult{0, []StepResult{}, encodeError(err)}
+	}
+
+	rw.conn.Do("RPUSH", "tasks", string(jsonRedisMsg))
+
 	reply, err := redis.Strings(rw.conn.Do("BLPOP", "replies-"+guid.String(), rw.timeoutInSeconds))
 
 	if err != nil {
@@ -54,7 +74,7 @@ func StartSlave(conn redis.Conn, delegate Worker) slave {
 func (slave slave) Close() error {
 	_, err := slave.conn.Do("RPUSH", "stop-"+slave.guid, true)
 	if err == nil {
-		_, err = slave.conn.Do("BLPOP", "stopped-"+slave.guid, DEFAULT_TIMEOUT)
+		_, err = slave.conn.Do("BLPOP", "stopped-"+slave.guid, DefaultTimeout)
 	}
 
 	logs.NewLogger("redis.slave").Infof("Redis slave shutting down, %v", err)
@@ -62,6 +82,8 @@ func (slave slave) Close() error {
 }
 
 func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
+	var redisMsg redisMessage
+
 	logger := logs.NewLogger("redis.slave")
 	logger.Info("Started slave")
 
@@ -78,16 +100,18 @@ func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
 		}
 
 		if err == nil {
-			parts := strings.SplitN(reply[1], " ", 3)
 
-			go func(experiment string, replyTo string, workerIndex string) {
-				i, _ := strconv.Atoi(workerIndex)
-				result := delegate.Time(experiment, i)
+			redisMsg.WorkloadContext = context.New()			
+
+			json.Unmarshal([]byte(reply[1]), &redisMsg)
+
+			go func(experiment string, replyTo string, workloadCtx context.Context) {
+				result := delegate.Time(experiment, workloadCtx)
 				var encoded []byte
 				encoded, err = json.Marshal(result)
 				logger.Debug("Completed slave task, replying")
 				conn.Do("RPUSH", replyTo, string(encoded))
-			}(parts[2], parts[0], parts[1])
+			}(redisMsg.Workload, redisMsg.Reply, redisMsg.WorkloadContext)
 		}
 
 		if err != nil {
--- a/benchmarker/redis_test.go
+++ b/benchmarker/redis_test.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/cloudfoundry-incubator/pat/context"
 	"github.com/cloudfoundry-incubator/pat/redis"
 	"github.com/cloudfoundry-incubator/pat/workloads"
 	. "github.com/onsi/ginkgo"
@@ -16,7 +17,8 @@ import (
 
 var _ = Describe("RedisWorker", func() {
 	var (
-		conn redis.Conn
+		conn        redis.Conn
+		workloadCtx context.Context
 	)
 
 	BeforeEach(func() {
@@ -24,6 +26,7 @@ var _ = Describe("RedisWorker", func() {
 		var err error
 		conn, err = redis.Connect("", 63798, "p4ssw0rd")
 		Ω(err).ShouldNot(HaveOccurred())
+		workloadCtx = context.New()
 	})
 
 	AfterEach(func() {
@@ -34,23 +37,24 @@ var _ = Describe("RedisWorker", func() {
 		Context("When no slaves are running", func() {
 			It("Times out after a specified time", func() {
 				worker := NewRedisWorkerWithTimeout(conn, 1)
+				workloadCtx.PutInt("workerIndex", 0)
 				worker.AddWorkloadStep(workloads.Step("timesout", func() error { time.Sleep(10 * time.Second); return nil }, ""))
-
 				result := make(chan error)
 				go func() {
-					result <- worker.Time("timesout", 0).Error
+					result <- worker.Time("timesout", workloadCtx).Error
 				}()
-
 				Eventually(result, 2).Should(Receive())
-			})			
+			})
 		})
 
 		Context("When a slave is running", func() {
 			var (
-				slave    io.Closer
-				delegate *LocalWorker
-				context  map[string]interface{}
-				wasCalledWithWorkerIndex int
+				slave                       io.Closer
+				delegate                    *LocalWorker
+				wasCalledWithWorkerIndex    int
+				wasCalledWithWorkerUsername string
+				wasCalledWithRandomKey      string
+				wasCalledWithBoolTypeKey    bool
 			)
 
 			JustBeforeEach(func() {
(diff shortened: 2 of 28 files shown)


benchmarker/redis.go
```diff
@@ -17,19 +16,41 @@ type rw struct {
 	timeoutInSeconds int
 }
 
-const DEFAULT_TIMEOUT = 60 * 5
+type RedisMessage struct { 
+  	Reply string
```
julz: this should be ReplyTo for clarity

benchmarker/redis.go
```diff
@@ -17,19 +16,41 @@ type rw struct {
...
+  	Workload string
+ 	WorkloadContext context.Context
+
+  }
+
+  type RedisMessageReply struct {
```
julz: go fmt?

benchmarker/redis.go
```diff
@@ -17,19 +16,41 @@ type rw struct {
...
+  	Reply string
+  	Workload string
+ 	WorkloadContext context.Context
+
+  }
+
+  type RedisMessageReply struct { 
+  	Reply string
+  	Workload string
+ 	WorkloadContext map[string]interface{}
```
julz: why does the reply have a context in it? does it go anywhere? either way, the type should be context.Context I think

benchmarker/redis.go
```diff
@@ -78,16 +102,17 @@ func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
 		}
 
 		if err == nil {
-			parts := strings.SplitN(reply[1], " ", 3)
 
-			go func(experiment string, replyTo string, workerIndex string) {
-				i, _ := strconv.Atoi(workerIndex)
-				result := delegate.Time(experiment, i)
+			json.Unmarshal([]byte(reply[1]), &redisMsg)
+			workloadCtx = context.NewWithContext(redisMsg.WorkloadContext)
```
julz: context.NewWithContext doesn't seem like it makes sense.. if you have a context, why do you need to new anything?

context/context.go
```diff
@@ -0,0 +1,105 @@
...
+	Clone() contextMap
+}
+
+type contextMap map[string]interface{}
+
+func New() Context {
+	return make(contextMap)
+}
+
+func NewWithContext(ctx map[string]interface{}) Context {
```
julz: this should be called NewFromMap, but I don't see why we need it.. what should have the right to create a context from a map? we create a context with .New, send it over the network with .MarshalJSON and decode it back to a  context with .UnmarshalJSON.. I don't see where in that process there's a need to construct a context from a map

context/context.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (c contextMap) MarshalJSON() ([]byte, error) {
+	return json.Marshal(map[string]interface{}(c))
+}
+
+func (c contextMap) Clone() contextMap {
+	var clone = make(contextMap)
+	for k, v := range c {
+    	clone[k] = v
```
julz: go fmt?

context/context.go
simonleung8: Couldn't get UnmarshalJSON() called, that is why I have it this way

julz: Let's fix it properly, we should be able to make UnmarshalJSON work, or if not understand exactly why we can't.

simonleung8: I updated the code.  Not using UnmarshalJSON() because we can't implement pointer receiver (c *ContextMap)UnmarshalJSON on an interface, without pointer it is always passed by value and I don't know how to get the unmarshaled context back into the caller.

benchmarker/redis.go
```diff
@@ -78,16 +100,18 @@ func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
 		}
 
 		if err == nil {
-			parts := strings.SplitN(reply[1], " ", 3)
 
-			go func(experiment string, replyTo string, workerIndex string) {
-				i, _ := strconv.Atoi(workerIndex)
-				result := delegate.Time(experiment, i)
+			redisMsg.WorkloadContext = context.New().(context.ContextMap)
```
julz: this cast shouldn't be needed, I double-checked locally and as long as the object you're unmarshalling in to is of the right runtime type, it should work

benchmarker/redis.go
```diff
@@ -17,19 +16,40 @@ type rw struct {
...
+type RedisMessage struct {
+	Reply           string
+	Workload        string
+	WorkloadContext context.Context
+}
+
+type RedisMessageReply struct {
+	Reply           string
+	Workload        string
+	WorkloadContext context.ContextMap
```
julz: this can just be context.Context, as long as you make sure that at runtime there's a ContextMap in this field (which there will be if you put the return value of context.New() in there)

benchmarker/redis.go
```diff
@@ -17,19 +16,40 @@ type rw struct {
 	timeoutInSeconds int
 }
 
-const DEFAULT_TIMEOUT = 60 * 5
+type RedisMessage struct {
```
julz: no reason these structs need to be exported

benchmarker/redis_test.go
```diff
@@ -126,7 +151,65 @@ var _ = Describe("RedisWorker", func() {
...
 					Ω(result.Steps[0].Duration.Seconds()).Should(BeNumerically("~", 1, 0.1))
 					Ω(result.Steps[1].Duration.Seconds()).Should(BeNumerically("~", 2, 0.1))
-				})				
+				})
+			})
+
+			Describe("Workload context map sending over Redis", func() {
+
+				AfterEach(func() {
+					//workloadCtx = context.NewWorkloadContent()
```
julz: lose?

context/context.go
```diff
@@ -0,0 +1,106 @@
...
+	GetInt64(k string) (int64, bool)
+	PutFloat64(k string, v float64)
+	GetFloat64(k string) (float64, bool)
+	PutBool(k string, v bool)
+	GetBool(k string) (bool, bool)
+	MarshalJSON() ([]byte, error)
+	Clone() ContextMap
+}
+
+type ContextMap map[string]interface{}
```
julz: no need to export this

context/context.go
```diff
@@ -0,0 +1,106 @@
...
+
+func New() Context {
+	return make(ContextMap)
+}
+
+func NewWithMap(ctx map[string]interface{}) Context {
```
julz: do we need this?

benchmarker/redis.go
simonleung8: Does your tests pass in /benchmarker when you lose the typecasting? Mine is failing.

benchmarker/redis.go
simonleung8: Isn't json.Unmarshal() requires the struct it is unmarshalling into exported?

julz: nope

simonleung8:
Code updated, I think the context map is now exactly how we want it to be.

benchmarker/redis_test.go
```diff
@@ -16,7 +17,8 @@ import (
 
 var _ = Describe("RedisWorker", func() {
 	var (
-		conn redis.Conn
+		conn        redis.Conn
+		workloadCtx = context.New()
```
julz: this initialisation should be in the BeforeEach to prevent interference between the tests

benchmarker/redis.go
```diff
@@ -78,16 +94,19 @@ func slaveLoop(conn redis.Conn, delegate Worker, handle string) {
 		}
 
 		if err == nil {
-			parts := strings.SplitN(reply[1], " ", 3)
 
-			go func(experiment string, replyTo string, workerIndex string) {
-				i, _ := strconv.Atoi(workerIndex)
-				result := delegate.Time(experiment, i)
+			c := context.New()
+			redisMsg.WorkloadContext = c
```
julz: this can just be `redisMsg.WorkloadContext = context.New()`

benchmarker/redis.go
```diff
@@ -17,19 +16,34 @@ type rw struct {
...
 	guid, _ := uuid.NewV4()
-	rw.conn.Do("RPUSH", "tasks", "replies-"+guid.String()+" "+strconv.Itoa(workerIndex)+" "+experiment)	
+	redisMsg := redisMessage{
+		Workload:        workload,
+		Reply:           "replies-" + guid.String(),
+		WorkloadContext: workloadCtx,
+	}
+
+	var jsonRedisMsg []byte
+	jsonRedisMsg, _ = json.Marshal(redisMsg)
```
julz: can we do something better with the error?

benchmarker/redis_test.go
```diff
@@ -59,12 +65,29 @@ var _ = Describe("RedisWorker", func() {
...
 
-				context = make(map[string]interface{})
-				delegate.AddWorkloadStep(workloads.StepWithContext("fooWithContext", func(ctx map[string]interface{}) error { context = ctx; ctx["a"] = 1; return nil }, ""))
-				delegate.AddWorkloadStep(workloads.StepWithContext("barWithContext", func(ctx map[string]interface{}) error { ctx["a"] = ctx["a"].(int) + 2; return nil }, ""))
-				delegate.AddWorkloadStep(workloads.StepWithContext("recordWorkerIndex", func(ctx map[string]interface{}) error { wasCalledWithWorkerIndex = ctx["workerIndex"].(int); return nil }, ""))
+				localContext = context.New()
+				delegate.AddWorkloadStep(workloads.StepWithContext("fooWithContext", func(ctx context.Context) error { localContext = ctx; ctx.PutInt("a", 1); return nil }, ""))
+				delegate.AddWorkloadStep(workloads.StepWithContext("barWithContext", func(ctx context.Context) error { a, _ := ctx.GetInt("a"); ctx.PutInt("a", a+2); return nil }, ""))
+				delegate.AddWorkloadStep(workloads.StepWithContext("recordWorkerIndex", func(ctx context.Context) error { wasCalledWithWorkerIndex, _ = ctx.GetInt("workerIndex"); return nil }, ""))
```
julz: there are a LOT of these lines now.. is there any way we can reduce the duplication a bit here?

context/context.go
```diff
@@ -0,0 +1,109 @@
...
+func (c contextMap) GetInt(k string) (int, bool) {
+	if c[k] == nil {
+		return 0, false
+	} else {
+		value, _ := strconv.ParseInt(c[k].(string), 0, 0)
+		return int(value), true
+	}
+}
+
+func (c contextMap) PutInt64(k string, v int64) {
```
julz: do we need GetInt/GetInt64 separately from GetInt?

simonleung8:
Added rest:api ui to this branch

ui/js/tests.js
```diff
@@ -203,6 +203,98 @@ describe("DOM elements manipulation", function(){
...
+  })
+
+  it("returns a list of selected commands, separated by commas", function(){
+    $("#workloadItems button:contains('gcf:push')").trigger("click")
+    $("#workloadItems button:contains('dummyWithErrors')").trigger("click")
+    $("#workloadItems button:contains('gcf:push')").trigger("click")
+
+    expect( WL.workloads() ).toBe("gcf:push,dummyWithErrors,gcf:push")
+
+    //clean up
```
julz: ?

ui/js/app.js
```diff
@@ -113,12 +113,25 @@ pat.view = function(experimentList, experiment) {
   this.workloadVisible = ko.observable(true)
   this.throughputVisible = ko.observable(false)
 
+  WL = new patWorkload(document.getElementById("workloadItems"), document.getElementById("selectedList"), document.getElementById("argumentList"));
```
julz: can we please not call this WL, it sounds like you're shouting. How about 'workloadList'?

ui/js/tests.js
```diff
@@ -5,7 +5,7 @@ describe("The view", function() {
   var throughputNode
 
   beforeEach(function() {
-    experiment = { run: function() {}, url: ko.observable(""), state: ko.observable(""), view: function() {}, csvUrl: ko.observable(""), config: { iterations: ko.observable(1), concurrency: ko.observable(1), interval: ko.observable(0), stop: ko.observable(0) } }
+    experiment = { run: function() {}, url: ko.observable(""), state: ko.observable(""), view: function() {}, csvUrl: ko.observable(""), config: { iterations: ko.observable(1), concurrency: ko.observable(1), interval: ko.observable(0), stop: ko.observable(0), cfTarget: ko.observable("http://xio.10.xx.xx.xx"), cfUsername: ko.observable("admin"), cfPassword: ko.observable("admin") } }
```
julz: the tests break for me unless I also add "workloads: ko.observable()" to this line ?

ui/js/tests.js
```diff
@@ -203,6 +203,98 @@ describe("DOM elements manipulation", function(){
 
 })
 
+describe("Workload List", function(){
```
julz: all of these tests fail if I run them on firefox, which probably means we have bugs on firefox

ui/js/app.js
```diff
@@ -113,12 +113,25 @@ pat.view = function(experimentList, experiment) {
...
+  WL.importKoBindingsVars("cfTarget", "targetHasError", "cfUsername", "usernameHasError", "cfPassword", "passwordHasError");
```
julz: is there any way we can avoid doing this? lots of coupling here

ui/js/workloadList.js
```diff
@@ -0,0 +1,252 @@
...
+	return {		
+		workloads: workloads,		
+		workloadsObservable: workloadsObservable,
+		importKoBindingsVars: importKoBindingsVars,
+		isArgumentError: isArgumentError,
+		workloadItems: workloadItems
+	};	
+
+	function drawArgumentInputs() {
+		for (var cmd in argumentItems) {
```
julz: instead of looping here, why not just use a [foreach binding](http://knockoutjs.com/documentation/foreach-binding.html). That way you can get rid of all this control flow logic and have the view be in the HTML, rather than hard-coded here. Also I think that way you don't need to do the weird importKoBindings thing above

simonleung8: The foreach binding is a good idea.  I probably don't have enough experience with knockout, but isn't that we still need some way to export the ko binding vars (cfTarget, cfUser, cfPass) from app.js to workloadList.js? How would the 'foreach' know what vars to use in data-bind="value: cfxxx"?

julz: if you use a binding then all the ko-related stuff will be in the html. So you'll just have a binding like  `<div data-bind="foreach: neededParameters"><input data-bind="style: {'hasError': hasErrors}; value: value"></div>`, and in app.js you'll export `neededParameters` (which workloadList will generate) as a `ko.observableArray()` which has a field for each parameter. So there should be no need for you to bind some specific name in the html, because in the html you're just looping over an observableArray that has proper elements in it, all of which have a 'value', 'hasErrors' and 'name' method that the loop can generically use

simonleung8:
Code updated:
- Using ko foreach binding for workload arguments, no more importing ko variables into workloadList.js
- Passes all tests in firefox
- better variables naming

ui/js/workloadList.js
```diff
@@ -0,0 +1,242 @@
...
+	var commandArgs = {};
+	var selectedWorkloads  = [];
+	var wlArgModel = argumentModel();
+
+	$(workloadNode).html('')
+	$(selectedNode).html('')
+
+	// draw workload buttons
+	for (var cmd in workloadItems) {
+		workloadBtns[cmd] = $("<button>", {
```
julz: why not do this in the html with a binding too?

ui/js/workloadList.js
```diff
@@ -0,0 +1,242 @@
...
+			return;
+		} 
+
+		var index = findIndexByElement(btn);
+		selectedWorkloads[index].remove();
+		selectedWorkloads.splice(index, 1);
+
+		if (!duplicatedCmd) {
+			//hide required arguments when command is removed				
+			workloadItems[btn.cmd].args.forEach(function(required) {
```
julz: instead of removing and adding, just bind the html to the list of current required parameters and recalculate when the list is changed

ui/js/workloadList.js
```diff
@@ -0,0 +1,242 @@
...
+		//add the required commands
+		workloadItems[cmd].requires.forEach(function(d){
+			if (findIndexByCommand(d) == -1) appendWorkloadBtnToDom(d)
+		})
+		appendWorkloadBtnToDom(cmd)
+	}
+
+	function appendWorkloadBtnToDom(cmd) {
+		var btn = {};				
+		btn = $("<button>", {
```
julz: -> binding; ideally there should be no html-related stuff in here; that will make it much easier to test everything, it'll make the javascript much simpler and clearer, it'll be much less likely to contain bugs _and_ it'll let us play with the UI if we want to without digging in to the javascript

simonleung8:
Code updated:
- All html is now rendered with foreach binding
- code clean up

julz:
lgtm

julz:
oh darn, can you rebase? I was just about to hit merge but github's complaining

simonleung8:
Rebased

PR #92, opened 2014-05-11
Fixed interval/stop to work with ramping concurrency

jberkhahn:
Changed runner.go so that a new schedule is created for each separate ExecuteConcurrently(). The changes to benchmarker.go are due to gofmt.

### DIFF
--- a/experiment/runner_test.go
+++ b/experiment/runner_test.go
@@ -34,7 +34,7 @@ var _ = Describe("ExperimentConfiguration and Sampler", func() {
 				sampler = &DummySampler{maxIterations, samples, iterationResults, workers, errors, sampleFunc}
 				return sampler
 			}
-			config = &RunnableExperiment{ExperimentConfiguration{5, []int{2}, 1*time.Second, 1, 3, worker, "push"}, executorFactory, samplerFactory}
+			config = &RunnableExperiment{ExperimentConfiguration{5, []int{2}, 1 * time.Second, 1, 3, worker, "push"}, executorFactory, samplerFactory}
 		})
 
 		It("Sends Samples from Sampler to the passed tracker function", func() {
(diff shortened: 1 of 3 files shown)


julz:
looks mostly good, but you have the 'go fmt' changes and the new code in one commit, which means I can't easily tell if you actually added a test for the fix.. if I revert the changes in runner.go, will runner_test.go fail?

jberkhahn:
Yes, the tests will fail. I didn't actually add any tests, merely  changed existing ones, as the fix is a glorified private wrapper function.

julz:
I see.. they won't compile at all if I revert runner.go, but if I add a `Start()` method that just returns the existing channel (and otherwise keep the implementation the same as it was) they won't fail. Let's add a test that shows that if you call `.Start()` on a concurrencySchedule it starts again from 0 (which is the behaviour we're relying on).

jberkhahn:
Ok, changed start to be a private method as there's really no reason for it to be exported, and added a test that checks that calling start repeatedly on a schedule restarts it.

experiment/runner_test.go
```diff
@@ -297,10 +299,42 @@ var _ = Describe("ExperimentConfiguration and Sampler", func() {
...
+					Ω(<-schedule).ShouldNot(BeNil())
+				}
 				Ω(schedule).Should(BeClosed())
 			})
+
+			Context("Repeated scheduling", func() {
+				It("creates a new schedule each time start() is called", func() {
+					scheduler := linearSchedule(1, 3, 3*time.Second)
+					schedule := scheduler.start()
+					for i := 0; i < 1; i++ {
```
julz: I might be being dumb.. but what's the point of this loop? ditto the loop on 325..

jberkhahn:
Woops, that was a remnant of an earlier version of the test. Removed.

PR #93, opened 2014-05-12
Colored command Barchart

simonleung8:
The barchart now have color coded blocks to show the total run time for each individual command in an iteration

### DIFF
--- a/ui/js/bar.js
+++ b/ui/js/bar.js
@@ -1,6 +1,7 @@
 d3_workload = function() {
   const barWidth = 30,
-        clipMarginRight = 0;
+        clipMarginRight = 0,
+        second = 1000000000;
 
   var margin = {top: 50, right: 40, bottom: 30, left: 30};
   var svgWidth, svgHeight, jqObj, d3Obj, drawArea;
@@ -20,6 +21,7 @@ d3_workload = function() {
     svgHeight = jqObj.height() - margin.top - margin.bottom;
     x = d3.scale.linear().range([0, svgWidth], 1),
     y = d3.scale.linear().range([0, svgHeight], 1);
+    color = d3.scale.category10();
 
     xAxis = d3.svg.axis()
       .scale(x)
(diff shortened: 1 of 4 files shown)


ui/js/bar.js
```diff
@@ -102,41 +103,72 @@ d3_workload = function() {
...
 
-    labels.transition()
-      .attr("x", function(d, i) { return (x(i) + (barWidth / 2)) })
-      .attr("y", svgHeight + 3 );
+    //draw command blocks
+    bars.transition()
+      .attr("x", function(d, i) { return x(d.iteration) } )
+      .attr("y", function(d) { return y((d.start + d.duration)/ second) } )
+      .attr("height", function(d) { return (svgHeight - y((d.duration) / second)) } )
+      .style("fill", function(d) { return color(d.cmd) } )
```
julz: is there a test for the colouring?

ui/js/bar.js
```diff
@@ -102,41 +103,72 @@ d3_workload = function() {
...
-    bars.transition()
-      .attr("x", function(d, i) { return x(i) })
-      .attr("y", function(d) { return y(d.LastResult / second) } )
-      .attr("height", function(d) { return (svgHeight - y(d.LastResult / second)) } );
+    color.domain(data.map(function (d) { return d.cmd; }));
 
-    labels.transition()
-      .attr("x", function(d, i) { return (x(i) + (barWidth / 2)) })
-      .attr("y", svgHeight + 3 );
+    //draw command blocks
```
julz: maybe pull out some of this stuff in to functions? the comments are probably a clue which things should be refactored..

ui/js/bar.js
```diff
@@ -170,25 +203,70 @@ d3_workload = function() {
...
+        
+      errorCount += 1;
+      $(this).tooltip({
+        "placement": "right",
+        "container": "body"
+      });
+    })
+  }
+
+  function flattenCommands(data) {
```
julz: I don't really understand what this function does, and I read it a few times.. could we pull it out to a util object and put a test around it so it's clear how it works?

ui/js/bar.js
```diff
@@ -170,25 +203,70 @@ d3_workload = function() {
...
         });
       } else {
-        d3.select(this).classed("error", false);
+        d3.select(this).classed("error", false)
+          .style("stroke", "none");
       }
     })
   }
 
+  function tooltipCommands() {
```
julz: what about them? can I buy a verb?

ui/js/bar.js
```diff
@@ -170,25 +203,70 @@ d3_workload = function() {
...
+    var bars = d3.select(drawArea).selectAll("rect.iterations");
     bars.each(function(d,i) {
       if (d.TotalErrors > errorCount) {
         var bar = d3.select(this);     
         bar.classed("error", true)
           .attr("data-toggle","tooltip")
-          .attr("title", "Error: " + d.LastError);
-          
+          .attr("title", "Error: " + d.LastError )
+          .style("stroke-width", "5px")
```
julz: maybe some of this could go into the css (so we have less code and I don't feel weird about us not testing this stuff)

simonleung8:
updated with some code refactoring and new tests

ui/js/tests.js
```diff
@@ -503,11 +502,83 @@ describe("Bar chart", function() {
...
+                  "dummy1":{ "LastTime": 1 * sec },                  
+                  "dummy2":{ "LastTime": 3 * sec }, 
+                  "dummy3":{ "LastTime": 2 * sec }}, 
+                "LastResult" : 6 * sec }];
+    
+    var cmdData = d3_workload.flattenWorkloadCommandsJSON(data)
+    
+    expect(cmdData[0]).toEqual({"cmd" : "dummy1", "start" : 0, "duration" : 1 * sec, "iteration" : 0})
+    expect(cmdData[1]).toEqual({"cmd" : "dummy2", "start" : 1 * sec, "duration" : 3 * sec, "iteration" : 0})
+    expect(cmdData[2]).toEqual({"cmd" : "dummy3", "start" : 4 * sec, "duration" : 2 * sec, "iteration" : 0})
```
julz: so I finally worked out after staring at this for a while why "start" is 4 seconds on this line. It's because the previous commands add up to 4, so we assume they executed in order and so the third command started after the second. Which is probably true, I think, but I'm not sure it's a great idea to do that in the javascript rather than having the server record the start times of the experiments.

I think what this method is REALLY doing is converting Commands (which is not really JSON at this point, it's just a hash, but anyway) in to coordinates for a bar graph. In which case let's call it `toBarGraph(commands)` and have it return e.g. '{label: "dummy3", y: 4, height: 2}'. Then we can have the bar chart be nice and generic and reusable and nothing to do with commands, and this function will just be mapping to that format. What do you think?

Apart from that this all looks good now

simonleung8:
Updated with your suggestion

ui/js/tests.js
julz: did you rerun the tests? surely this should fail now??..

simonleung8:
Rebased

julz:
did you rerun the tests? shouldn't the "it has a utility function to turn 'data' into array containing" fail now?? (it's still calling `flattenWorkloadCommandsJSON()` and we changed the method name to `toBarGraph()`..)

simonleung8:
Totally forgot to update the test..  It is updated now.

julz:
Just realised we don't have the jasmine tests running as part of travis.. we could really do with adding those, do you think you could try to add them to travis.yml as part of this PR so we get a green light on it? we already have phantomjs set up to run the feature tests so I _think_ it should be quite easy to add; otherwise we'll have to pull and run the javascript tests manually when reviewing PRs which'd be a big pain

simonleung8:
Jasmine tests are now part of travis.  Had to modify the tests a little to satisfy phantomjs, I guess phantomjs's javascript support is a tiny different.

Total Jasmine tests: 69
Total passed: 69

julz:
_Awesome_, nice, think you typo-ed the commit message though when you squashed..

simonleung8:
Fixed last commit message

PR #94, opened 2014-05-19
More intuiative and nicer workload input for user

simonleung8:
Small update for:
- Dividing workload input (workload + arguments) and other inputs (iteration, concurrency interval etc ...) into left and right panels.
- Only shows the captions 'Selected Workload' and 'Argument Input' when necessary to avoid confusing users.

### DIFF
--- a/ui/js/workloadModels.js
+++ b/ui/js/workloadModels.js
@@ -224,10 +224,27 @@ patWorkload = function(){
 	var selectedModel = ko.observableArray([])
 	var argumentModel = ko.observableArray([cfTarget, cfUser, cfPass]);
 
+	var showSelectedCaption = ko.computed(function(){
+		if (selectedModel().length > 0) {
+			return true
+		} else {
+			return false
+		}
+	})
+
+	var showArgumentCaption = ko.computed(function(){			
+		for (var i = 0; i < argumentModel().length; i ++) {
+			if (argumentModel()[i].display() != "none") return true
+		}
+		return false
+	})
+
 	return {
 		workloads: workloadStr,
 		itemModel: itemModel,
 		selectedModel: selectedModel,
+		shouldShowSelectedCaption: showSelectedCaption,
+		shouldShowArgumentCaption: showArgumentCaption,		
 		argumentModel: argumentModel,
 		cfTargetHasErr: cfTarget.errCheckFn,
 		cfUserHasErr: cfUser.errCheckFn,
--- a/ui/css/app.css
+++ b/ui/css/app.css
@@ -34,6 +34,15 @@ tr, td {
   box-sizing:border-box;
 }
 
+.inputCaption{
+  background: #ccc;
+  display: block;
+  padding: 2px 5px;
+  color: #999;
+  font-size: 10pt;
+  text-shadow: 0 1px 0 #eee;
+}
+
 .main-tooltip {
   position:absolute;
   right:15px;
--- a/ui/index.html
+++ b/ui/index.html
@@ -79,78 +79,78 @@ body { padding: 8px; }
     </div>
   </div>
 
-  <div class="modal" id="experimentPopup" tabindex="-1" role="dialog" data-dismiss="modal" aria-labelledby="experimentPopupLabel" aria-hidden="true" >
+  <div class="modal" id="experimentPopup" tabindex="-1" role="dialog" aria-labelledby="experimentPopupLabel" aria-hidden="true" >
     <div class="modal-dialog" style="width: 90%; max-width: 900px;">
       <div class="modal-content" style="background:rgba(255,255,255,0.75);">
         <div class="modal-header">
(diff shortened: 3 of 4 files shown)


ui/js/workloadModels.js
```diff
@@ -89,6 +89,13 @@ patWorkload = function(){
 		requiredBy: ["rest:login"]
 	}
 
+	// UI Control Model
+	var uiControl =
```
julz: a) why the sub-object here? does it add anything? (honest question, if there's a good reason we can keep it, but I didn't immediately see it)
b) is this the most useless comment ever? ;-) :-) Why don't we just pull stuff in to separate files/objects rather than using comments to try to split one object in to several? If we're worried about performance we can pre-process and minify, let's not make our code a mess just because javascript

julz:
General thoughts (and this is me slightly taking my programmer hat off and stomping right in to Andy's territory, sorry!): I'm not convinced this is getting more intuitive. I just had a play and frankly it's quite a bit less intuitive than it used to be. Now in a way that's understandable because there's a lot more functionality (and it's really really cool stuff, btw!). But for the simple use case of "run a pre-existing experiment" I'm worried we've added a lot of overhead now. 

For example all the controls that really matter for most experiments (iterations/concurrency etc.) are now squashed together in a separate column, while loads of space is given to creating the workload script (which will not normally need to be edited, since you probably just want to use an existing one like login and push a few times).

Medium term, I think ideally we should have the main interface be a drop-down list like before of pre-made experiments, and then an 'add experiment' button that would present the interface you have here (but that's another story because then we need to save and restore experiment definitions, which we should add to the backlog). 

For now, how about we make the main interface the drop down, and then have an 'advanced' button that swaps to the interface you have here. That way we keep the initial "I want to try this out quickly and run an experiment" interface nice and simple and avoid new users running, screaming, away

/cc @bravera - thoughts?

ui/js/workloadModels.js
simonleung8: We could definitely have this object embedded into the 2 existing workload and argument models, but on the other hand, we certainly will add more functionalities, and there will be the need of further customization of the UI, so imo having the extra object for UI control make sense.

julz: I'd really like us to try to simplify this down a bit, it doesn't help us to have a sub-object here, I think you could replace all this stuff by just making showSelectedCaption be a ko.computed that used the other values to calculate whether the caption should be shown. 

I really don't like clients needing to do `workloadlist.uimodel.showSelectedCaption` rather than just `workloadList.shouldShowSelectedCaption`, it exposes the inner structure of workloadList for no real reason. If we later need a separate object (for reasons I can't immediately think of, but who knows) that method should still exist and delegate rather than exposing the inner objects to the client of this class.

ui/js/workloadModels.js
```diff
@@ -223,11 +252,13 @@ patWorkload = function(){
 	var itemModel = ko.observableArray([restTarget, restLogin, restPush, gcfPush, dummy, dummyWithErrors])
 	var selectedModel = ko.observableArray([])
 	var argumentModel = ko.observableArray([cfTarget, cfUser, cfPass]);
+	var uiControlModel = ko.observable(uiControl)
 
 	return {
 		workloads: workloadStr,
 		itemModel: itemModel,
 		selectedModel: selectedModel,
```
julz, line 245: do we particularly need to suffix these things with 'model'? isn't it just workloadList.selection, workloadList.arguments etc. (workloadList is the model object, not its properties)

simonleung8:
I agree with you having a dropdown for simple workload and have the advanced configuration hidden is a good idea.  I think I will have the feature added in another PR.  The main goal for this one is to hide the unneeded captions so it doesn't confuse users. I will clean up the code and implement some of your suggestions in this one.

julz:
Ok, sounds good to me

simonleung8:
Updated, no more uiControl object, using ko.computed() instead

ui/js/workloadModels.js
```diff
@@ -224,10 +224,27 @@ patWorkload = function(){
...
+			if (argumentModel()[i].display() != "none") return true
+		}
+		return false
+	})
+
 	return {
 		workloads: workloadStr,
 		itemModel: itemModel,
 		selectedModel: selectedModel,
+		shouldShowSelectedCaption: showSelectedCaption,
```
julz, line 246: see my comments in the other PR about this pattern of double-exporting in the return statement which means you have to go look here to find what the function is called, and then go find what it's actually mapped to. Let's lose this and use an 'exports' object instead (or similar). 

apart from that though this looks way better, going to merge but can we get rid of the double-returns as part of the other PR please?

julz:
merging but please let's get rid of the double-returns as part of the other open PR, also let's make sure to come back asap and add a simple drop down with this as an advanced option, I really hate the new user experience right now

PR #95, opened 2014-05-20
Allow user to configure 'space' in web ui

simonleung8:
Adding support for configurable -rest:space in Web UI, it is now an argument field with the login command

### DIFF
--- a/cmdline/cmdline.go
+++ b/cmdline/cmdline.go
@@ -26,6 +26,10 @@ var params = struct {
 	workload            string
 	interval            int
 	stop                int
+	restUser            string
+	restPass            string
+	restTarget          string
+	restSpace           string
 }{}
 
 func InitCommandLineFlags(config config.Config) {
@@ -38,12 +42,19 @@ func InitCommandLineFlags(config config.Config) {
 	config.IntVar(&params.interval, "interval", 0, "repeat a workload every n seconds, to be used with -stop")
 	config.IntVar(&params.stop, "stop", 0, "repeat a repeating interval until n seconds, to be used with -interval")
 	config.BoolVar(&params.listWorkloads, "list-workloads", false, "Lists the available workloads")
+	config.StringVar(&params.restTarget, "rest:target", "", "the target for the REST api")
+	config.StringVar(&params.restUser, "rest:username", "", "username for REST api")
+	config.StringVar(&params.restPass, "rest:password", "", "password for REST api")
+	config.StringVar(&params.restSpace, "rest:space", "dev", "space to target for REST api")
 	benchmarker.DescribeParameters(config)
 	store.DescribeParameters(config)
 }
 
 func RunCommandLine() error {
 	params.workload = strings.Replace(params.workload, " ", "", -1)
+
+	workloadContext := context.New()
+	workloads.PopulateRestContext(params.restTarget, params.restUser, params.restPass, params.restSpace, workloadContext)
 
 	return WithConfiguredWorkerAndSlaves(func(worker benchmarker.Worker) error {
 		return validateParameters(worker, func() error {
@@ -60,8 +71,6 @@ func RunCommandLine() error {
 						display(params.concurrency, params.iterations, params.interval, params.stop, params.concurrencyStepTime, s)
 					})
 				}
-
-				workloadContext := context.New()
 
 				lab.RunWithHandlers(
 					NewRunnableExperiment(
--- a/context/context.go
+++ b/context/context.go
@@ -2,11 +2,13 @@ package context
 
 import (
 	"encoding/json"
+	"reflect"
 	"strconv"
 )
 
 type Context interface {
 	PutString(k string, v string)
+	PutStringPtr(k string, v *string)
 	GetString(k string) (string, bool)
 	PutInt(k string, v int)
 	GetInt(k string) (int, bool)
@@ -30,9 +32,15 @@ func (c contextMap) PutString(k string, v string) {
 	c[k] = v
 }
 
+func (c contextMap) PutStringPtr(k string, ptr *string) {
+	c[k] = ptr
+}
+
 func (c contextMap) GetString(k string) (string, bool) {
 	if c[k] == nil {
 		return "", false
+	} else if reflect.TypeOf(c[k]).Kind() == reflect.Ptr {
+		return *c[k].(*string), true
 	} else {
 		return c[k].(string), true
 	}
--- a/context/context.go
+++ b/context/context.go
@@ -2,13 +2,11 @@ package context
 
 import (
 	"encoding/json"
-	"reflect"
 	"strconv"
 )
 
 type Context interface {
 	PutString(k string, v string)
-	PutStringPtr(k string, v *string)
 	GetString(k string) (string, bool)
 	PutInt(k string, v int)
 	GetInt(k string) (int, bool)
@@ -32,15 +30,9 @@ func (c contextMap) PutString(k string, v string) {
 	c[k] = v
 }
 
-func (c contextMap) PutStringPtr(k string, ptr *string) {
-	c[k] = ptr
-}
-
 func (c contextMap) GetString(k string) (string, bool) {
 	if c[k] == nil {
 		return "", false
-	} else if reflect.TypeOf(c[k]).Kind() == reflect.Ptr {
-		return *c[k].(*string), true
 	} else {
 		return c[k].(string), true
 	}
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ func main() {
 	logs.InitCommandLineFlags(flags)
 	cmdline.InitCommandLineFlags(flags)
 	server.InitCommandLineFlags(flags)
+
 	err := flags.Parse(os.Args[1:])
 	if err != nil {
 		fmt.Println(err)
--- a/main.go
+++ b/main.go
@@ -18,7 +18,6 @@ func main() {
 	logs.InitCommandLineFlags(flags)
 	cmdline.InitCommandLineFlags(flags)
 	server.InitCommandLineFlags(flags)
-
 	err := flags.Parse(os.Args[1:])
 	if err != nil {
 		fmt.Println(err)
--- a/server/server.go
+++ b/server/server.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/cloudfoundry-incubator/pat/benchmarker"
@@ -134,10 +133,11 @@ func (ctx *serverContext) handlePush(w http.ResponseWriter, r *http.Request) (in
 		stop = 0
 	}
 
-	cfTarget := trimSpaces(r.FormValue("cfTarget"))
-	cfUsername := trimSpaces(r.FormValue("cfUsername"))
-	workload := trimSpaces(r.FormValue("workload"))
-	cfPassword := trimSpaces(r.FormValue("cfPassword"))
+	cfTarget := r.FormValue("cfTarget")
+	cfUsername := r.FormValue("cfUsername")
+	workload := r.FormValue("workload")
+	cfPassword := r.FormValue("cfPassword")
+	cfSpace := r.FormValue("cfSpace")
 
 	if workload == "" {
 		workload = "gcf:push"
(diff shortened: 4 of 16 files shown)


server/server.go
```diff
@@ -138,6 +138,7 @@ func (ctx *serverContext) handlePush(w http.ResponseWriter, r *http.Request) (in
 	cfUsername := trimSpaces(r.FormValue("cfUsername"))
 	workload := trimSpaces(r.FormValue("workload"))
 	cfPassword := trimSpaces(r.FormValue("cfPassword"))
```
julz: seriously, why are we trimming any of these values? does any other web service trim spaces on input parameters? isn't <space> a valid value in a password meaning you made anyone with a password at the start/end of their password unable to log in?.. please let's lose the trimming?

server/server.go
```diff
@@ -148,6 +149,7 @@ func (ctx *serverContext) handlePush(w http.ResponseWriter, r *http.Request) (in
 	workloadContext.PutString("cfTarget", cfTarget)
 	workloadContext.PutString("cfUsername", cfUsername)
 	workloadContext.PutString("cfPassword", cfPassword)
+	workloadContext.PutString("cfSpace", cfSpace)
```
julz: just a thought, but maybe we should do this now rather than later; it'd be better to do something like "cf.space" rather than "cfSpace" if we want to namespace things, camel casing is harder to parse out etc.

ui/js/workloadModels.js
```diff
@@ -218,11 +233,12 @@ patWorkload = function(){
 	cfTarget.errCheckFn = ko.computed(function(){ return isTargetError(); })
 	cfUser.errCheckFn = ko.computed(function(){ return isUsernameError(); })
 	cfPass.errCheckFn = ko.computed(function(){ return isPasswordError(); })
+	cfSpace.errCheckFn = ko.computed(function(){ return isSpaceError(); })
```
julz: is there any point to the funcs-in-funcs-in-funcs here? why not just `cfSpace.errCheckFn = ko.computed( isSpaceError )`? in fact what's the point of setting cfSpace.errCheckFn here at all only to re-expose it later in the return value? why not just have the return statement be:

```
 return {
    ..
   cfSpaceHasError: ko.computed(spaceHasError)
   ..
 }
```

and lose all this stuff. this whole file could use some simplifying down, there's no need to be setting the same value 3 times in different objects. How about we just define an 'exports' object at the top of this object and then directly place the public function on that object (`exports.spaceHasError = ko.computed(function() { ... })`), and then return it at the bottom (`return exports`). Then we don't need the duplication in the return block at all.

ui/js/workloadModels.js
```diff
@@ -232,10 +248,12 @@ patWorkload = function(){
 		cfTargetHasErr: cfTarget.errCheckFn,
 		cfUserHasErr: cfUser.errCheckFn,
 		cfPassHasErr: cfPass.errCheckFn,
+		cfSpaceHasErr: cfSpace.errCheckFn,
 		worklistHasError: worklistHasError,
 		cfTarget: argumentModel()[0].value,
 		cfUsername: argumentModel()[1].value,
-		cfPassword: argumentModel()[2].value
+		cfPassword: argumentModel()[2].value,
+		cfSpace: argumentModel()[3].value
```
julz: what's the point of the 'argumentModel' (which, really, let's just call 'arguments') when we then specifically hard-code which of the arguments is cfSpace/cfPassword etc?

workloads/rest.go
```diff
@@ -69,6 +69,11 @@ func (r *rest) Login(ctx context.Context) error {
 func (r *rest) targetSpace(ctx context.Context) error {
 	apiEndpoint, _ := ctx.GetString("apiEndpoint")
 
+	if _, exists := ctx.GetString("cfSpace"); exists { 
+		r.space_name, _ = ctx.GetString("cfSpace")
+	}
+	fmt.Println(r.space_name)
```
julz: -> log

simonleung8:
Updated code to implement your suggestions

workloads/rest.go
```diff
@@ -69,6 +69,10 @@ func (r *rest) Login(ctx context.Context) error {
 func (r *rest) targetSpace(ctx context.Context) error {
 	apiEndpoint, _ := ctx.GetString("apiEndpoint")
 
+	if _, exists := ctx.GetString("cf.space"); exists {
```
julz: is there a test for this?

simonleung8:
Updated:
- small refactor to make it clear that context map is used to populate rest struct
- tests

ui/js/workloadModels.js
```diff
@@ -209,23 +218,38 @@ patWorkload = function(){
...
 
-	cfTarget.errCheckFn = ko.computed(function(){ return isTargetError(); })
-	cfUser.errCheckFn = ko.computed(function(){ return isUsernameError(); })
-	cfPass.errCheckFn = ko.computed(function(){ return isPasswordError(); })
+	cfTarget.errCheckFn = ko.computed( isTargetError )
+ 	cfUser.errCheckFn = ko.computed( isUsernameError )
+ 	cfPass.errCheckFn = ko.computed( isPasswordError )
+ 	cfSpace.errCheckFn = ko.computed( isSpaceError )
+
+ 	var validate = {}
```
julz: I'm being a pedant, but this should actually be 'validation', though honestly Im not sure what the sub-object buys us here, it's more typing and it doesn't abstract anything

ui/js/tests.js
```diff
@@ -124,58 +124,71 @@ describe("Workload List", function(){
...
     return -1;
   }
 
   function findArgumentIndex (cmd, start, end) {
     start = (start >= 0)? start : 0;
-    end = ( end >= 0)? end : workloadList.argumentModel().length;
+    end = ( end >= 0)? end : workloadList.argument().length;
     for (var i = start; i < end; i++) {      
-      if (workloadList.argumentModel()[i].forCmd == cmd) return i;
+      if (workloadList.argument()[i].forCmd == cmd) return i;
```
julz: argument -> arguments

ui/js/app.js
```diff
@@ -138,7 +139,7 @@ pat.view = function(experimentList, experiment) {
   this.numIntervalHasError = ko.computed(function() { return experiment.config.interval() < 0 })
   this.numStop = experiment.config.stop
   this.numStopHasError = ko.computed(function() { return experiment.config.stop() < 0 })
-  this.formHasNoErrors = ko.computed(function() { return ! ( this.workloadModels.worklistHasError() | this.workloadModels.cfTargetHasErr() | this.workloadModels.cfUserHasErr() | this.workloadModels.cfPassHasErr() | this.numIterationsHasError() | this.numConcurrentHasError() | this.numIntervalHasError() | this.numStopHasError() ) }, this)
+  this.formHasNoErrors = ko.computed(function() { return ! ( this.workloadModels.validate.WorkloadHasError() | this.workloadModels.validate.TargetHasError() | this.workloadModels.validate.UserHasError() | this.workloadModels.validate.PassHasError() | this.workloadModels.validate.SpaceHasError() | this.numIterationsHasError() | this.numConcurrentHasError() | this.numIntervalHasError() | this.numStopHasError() ) }, this)
```
julz: let's move this in to the workloadModel, i,e, just have a single `this.workloadModels.hasErrors()`, the workloadModel is the thing that knows if these subfields have errors or not, the outer object shouldn't need to check all of the inner fields individually to determine that

workloads/rest.go
```diff
@@ -240,3 +241,25 @@ func (r *rest) oauthInputs(username string, password string) url.Values {
...
+	if _, exists := ctx.GetString("cf.username"); exists {
+		r.username, _ = ctx.GetString("cf.username")
+	}
+
+	if _, exists := ctx.GetString("cf.password"); exists {
+		r.password, _ = ctx.GetString("cf.password")
+	}
+}
+
+func (r *rest) getTargetFromCtx(ctx context.Context) {
```
julz: these look like 'getter' functions, but they actually set a property on the instance... which is unexpected, to say the least; let's either a) look the values up in the context consistently throughout (and initialise the context from the cmd line parameters at startup for the cmdline case), or b) look the values up in the instance consistently throughout (and have each experiment use its own RestWorkload object, with its own parameters, rather than using the context); we shouldn't need all this confusion keeping the two in sync, we should just pick one or the other and stick to it

simonleung8:
Code upadated:
- context map gets initialized at main.go and gets passed onto server and cmdline
- rest.go DescribeParameters() populates contextmap
- rest.go only uses context map

main.go
```diff
@@ -26,9 +30,9 @@ func main() {
 
 	if useServer == true {
 		logs.NewLogger("main").Info("Starting in server mode")
-		server.Serve()
+		server.Serve(workloadContext)
```
julz: This is probably a pretty bad idea. We now have one context object that's getting reused by all experiments, so cruft from the previous experiment can affect the next one.

cmdline/cmdline.go
```diff
@@ -28,7 +28,7 @@ var params = struct {
 	stop                int
 }{}
 
-func InitCommandLineFlags(config config.Config) {
+func InitCommandLineFlags(config config.Config, ctx context.Context) {
```
julz: hmmm... I see what you're thinking here, but a) InitCommandLineFlags is for initting command line flags, not doing this  b) it seems odd for cmdline to know about the context, we're spraying this concept around quite a lot of layers now; wouldn't it be a lot nicer for us to just create a new RestWorkload before the experiment and pass in the initial parameters to its constructor; then it can decide to stash them in a context or whatever as needed?

server/server.go
```diff
@@ -145,9 +146,10 @@ func (ctx *serverContext) handlePush(w http.ResponseWriter, r *http.Request) (in
...
 
-	workloadContext.PutString("cfTarget", cfTarget)
-	workloadContext.PutString("cfUsername", cfUsername)
-	workloadContext.PutString("cfPassword", cfPassword)
+	workloadContext.PutString("rest:target", cfTarget)
+	workloadContext.PutString("rest:username", cfUsername)
+	workloadContext.PutString("rest:password", cfPassword)
+	workloadContext.PutString("rest:space", cfSpace)
 
 	experiment, _ := ctx.lab.Run(
```
julz, line 145: how about we just create a new worker here, and pass in to the constructor any parameters it should use; then the worker can be the thing that knows about the context, and since we have a new worker (and therefore context) each time, we're thread safe and state from previous experiments doesn't affect new ones

context/context.go
```diff
@@ -30,9 +32,15 @@ func (c contextMap) PutString(k string, v string) {
 	c[k] = v
 }
 
+func (c contextMap) PutStringPtr(k string, ptr *string) {
```
julz: I really don't like storing string pointer in the context, we can't possibly ship a pointer over the network, so this is going to be confusing with things like redis

simonleung8:
Updated.
context map instead of flags are used to bring arguments into rest.go

cmdline/cmdline.go
```diff
@@ -38,13 +42,20 @@ func InitCommandLineFlags(config config.Config) {
...
 	benchmarker.DescribeParameters(config)
 	store.DescribeParameters(config)
 }
 
-func RunCommandLine() error {
+func RunCommandLine(config config.Config) error {
 	params.workload = strings.Replace(params.workload, " ", "", -1)
 
+	workloadContext := context.New()
+	workloads.PopulateRestContext(config.ParseArgsForContext(), workloadContext)
```
julz: instead of this, why not remove the dependency of the config object and just do something explicit like `workloads.PopulateRestContext(params.restTarget, params.restUser, ..)` ?

config/config.go
```diff
@@ -118,3 +119,11 @@ func (f *f) ParseConfig(path string) error {
 
 	return nil
 }
+
+func (f *f) ParseArgsForContext() (argsMap map[string]string) {
```
julz: there's no need for this, config parses stuff into variables, we don't need a special function to create a map

context/context.go
julz: ^^

server/server.go
```diff
@@ -134,20 +134,13 @@ func (ctx *serverContext) handlePush(w http.ResponseWriter, r *http.Request) (in
...
 	if workload == "" {
 		workload = "gcf:push"
 	}
 
 	workloadContext := context.New()
-
-	workloadContext.PutString("cfTarget", cfTarget)
-	workloadContext.PutString("cfUsername", cfUsername)
-	workloadContext.PutString("cfPassword", cfPassword)
+	workloads.PopulateRestContext(populateArgumentMap(r), workloadContext)
```
julz: lets use explicit parameters rather than a map

workloads/rest_test.go
```diff
@@ -138,6 +145,15 @@ var _ = Describe("Rest", func() {
 						Ω(data.(url.Values)["grant_type"]).Should(Equal([]string{"password"}))
 					})
 
+					It("allows overridding username and password in context map", func() {
```
julz: now that all the logic for dealing with the context is inside rest.go, we should test only in those terms; i.e. if I call PopulateRestContext(..) with these parameters, then it has this effect (i.e. we don't expect users to directly use the context, so the tests should talk only in terms of the functions the user of this module is expected to use - PopulateRestContext, not PutString)

simonleung8:
Code updated.

julz:
looks good :)

PR #96, opened 2014-05-21
Fixed error reporting to so errors actually show up

jberkhahn:
Ok, this change includes a couple other quality-of-life improvements that I stumbled on while coding it up. Namely, the change to server_test.go to use TempDir() is necessary because the hardcoded temp file it was previously using was causing a test to fail due to lingering test files from old test runs that were in the old format. (also, it would most likely fail on a non-unix system). New version uses a system-safe method and now cleans up after itself. 
The change to server.go itself is because previously, the error was being upload as a Golang error object, which javascript doesn't know what to do with. I've created a generic method that takes a struct and converts it to a string/interface map, which should hopefully be generic enough to take anything you throw at it and convert it to something javascript can read. 

Bunch of gofmt changes to the files I touched.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -160,14 +160,13 @@ func (ctx *serverContext) handlePush(w http.ResponseWriter, r *http.Request) (in
 func (ctx *serverContext) handleGetExperiment(w http.ResponseWriter, r *http.Request) (interface{}, error) {
 	name := mux.Vars(r)["name"]
 	data, err := ctx.lab.GetData(name)
-
 	if len(data) == 0 {
 		// Ensure empty array is encoded as [] rather than null
 		/// see https://groups.google.com/forum/#!topic/golang-nuts/gOHbOk8DsFw
 		data = []*Sample{}
 	}
-
 	return &listResponse{data}, err
+
 }
 
 func csvHandler(fn func(http.ResponseWriter, *http.Request) (interface{}, error)) http.HandlerFunc {
--- a/store/csv.go
+++ b/store/csv.go
@@ -69,7 +69,7 @@ func (self *csvFile) Write(samples <-chan *experiment.Sample) {
 	var body []string
 	w := csv.NewWriter(f)
 
-	header = []string{"Average", "TotalTime", "Total", "TotalErrors", "TotalWorkers", "LastResult", "WorstResult", "NinetyfifthPercentile", "WallTime", "Type"}
+	header = []string{"Average", "TotalTime", "Total", "TotalErrors", "LastError", "TotalWorkers", "LastResult", "WorstResult", "NinetyfifthPercentile", "WallTime", "Type"}
 	for _, k := range self.commands {
 		header = append(header, "Commands|"+k+"|Count",
 			"Commands|"+k+"|Throughput",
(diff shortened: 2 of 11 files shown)


server/server.go
```diff
@@ -160,23 +161,27 @@ func (ctx *serverContext) handlePush(w http.ResponseWriter, r *http.Request) (in
...
 	if len(data) == 0 {
 		// Ensure empty array is encoded as [] rather than null
 		/// see https://groups.google.com/forum/#!topic/golang-nuts/gOHbOk8DsFw
 		data = []*Sample{}
 	}
 
-	return &listResponse{data}, err
+	experiments := make([]map[string]interface{}, 0)
+	for _, d := range data {
+		json := structToMap(*d)
```
julz: why? is this just so that you can convert the Error field so it gets encoded? a) if so then the method shouldn't be called structToMap which hides the actual intent of the method  b) this is a bad way of doing this, just use EncodableError to store the error and it'll get encoded properly (and you can always have EncodableError implement MarshalJSON to customise that encoding if you need to)

julz:
good but should use EncodableError or MarshalJSON rather than structToMap

jberkhahn:
Ok, got rid of the map converter and added custom json marshling for encodable errors. I had to export the encodable error type from within benchmarker, dunno how I feel about that.

store/csv.go
```diff
@@ -141,11 +144,18 @@ func (self *csvFile) GetData() (samples []*experiment.Sample, err error) {
 			sample.TotalTime, err = duration(d[1])
 			sample.Total, err = i64(d[2])
 			sample.TotalErrors, err = strconv.Atoi(d[3])
-			sample.TotalWorkers, err = strconv.Atoi(d[4])
-			sample.LastResult, err = duration(d[5])
-			sample.WorstResult, err = duration(d[6])
-			sample.NinetyfifthPercentile, err = duration(d[7])
-			sample.WallTime, err = duration(d[8])
+			sample.LastError = func(err string) error {
```
julz: surely we can do better than this?

server/server.go
```diff
@@ -160,14 +160,16 @@ func (ctx *serverContext) handlePush(w http.ResponseWriter, r *http.Request) (in
...
 	data, err := ctx.lab.GetData(name)
-
 	if len(data) == 0 {
 		// Ensure empty array is encoded as [] rather than null
 		/// see https://groups.google.com/forum/#!topic/golang-nuts/gOHbOk8DsFw
 		data = []*Sample{}
 	}
-
+	for _, d := range data {
+		d.LastError = benchmarker.EncodeError(d.LastError)
```
julz: instead of doing EncodeError here, could we encode the error inside benchmarker and avoid exporting EncodableError? (that way, anything we benchmark would already have an appropriate error and we wouldn't need this externally). Alternatively we could pull EncodableError in to a util module, but ideally it'd be better to just have the errors that come from benchmarker already be encodable.

store/csv.go
jberkhahn: I'm not really sure what you mean. Do you think it would be better to abstract this into it's own function?

server/server.go
jberkhahn: Except we already do: https://github.com/cloudfoundry-incubator/pat/blob/master/benchmarker/local.go#L26

The problem is we don't actually ever display results from a benchmarker anything. We display samples that have stats from a csv poured into them as "live" results. So, I can move the error encoding into store instead of server, but it's still not going to be in benchmarker.

julz: ok, that makes sense; I think store is a way better place to worry about whether errors are encodable or not than benchmarker, however we might be able to make this much simpler... why don't we just make the Error field in Sample/Store etc. a string (since that's actually all we care about at those points)? Then we can totally avoid all this silly EncodableError stuff..

store/csv.go
julz: worst case yes, that; but parsing a "<nil>" into nil is just painful... I think we should just make the LastError field a string in the first place and avoid all the silliness of encoding and decoding back and forth to errors when we only care about the message

jberkhahn:
Changed Sample to store the LastError as a string, rather than an error.

julz:
looks good

PR #97, opened 2014-05-22
fixed index error in percentile calculations fixed repeateveryuntil to be deterministc

jberkhahn:
There was an error in calculating the length of the array for calculating percentile when interval/stop aren't easily divisible. Fixed it so it should work with any (positive) numbers now.

There was another error in the RepeatEveryUntil method, namely there was a race condition between the interval and stop tickers that was causing the final workload that is supposed to occur when time=stop to only execute sometimes. Changed the method so it should always kick off the final iteration now.

### DIFF
--- a/benchmarker/benchmarker_test.go
+++ b/benchmarker/benchmarker_test.go
@@ -62,47 +62,55 @@ var _ = Describe("Benchmarker", func() {
 	})
 
 	Describe("RepeatEveryUntil", func() {
-		It("repeats a function at n seconds interval", func() {
+		It("repeats a function every interval seconds", func() {
 			start := time.Now()
 			var end time.Time
-			n := 2
-			Execute(RepeatEveryUntil(n, 3, func(context.Context) { end = time.Now() }, nil), workloadCtx)
+			interval := 2
+			Execute(RepeatEveryUntil(interval, 3, func(context.Context) { end = time.Now() }, nil), workloadCtx)
 			elapsed := end.Sub(start)
 			elapsed = (elapsed / time.Second)
-			Ω(int(elapsed)).Should(Equal(n))
+			Ω(int(elapsed)).Should(Equal(interval))
 		})
 
-		It("repeats a function at n seconds interval and stops at s second", func() {
+		It("repeats a function every interval seconds until stop seconds", func() {
 			var total int = 0
-			n := 2
-			s := 11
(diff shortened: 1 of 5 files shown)


experiment/runner.go
```diff
@@ -118,7 +117,7 @@ func (config *RunnableExperiment) Run(tracker func(<-chan *Sample), workloadCtx
 	done := make(chan bool)
 	maxIterations := config.Iterations
 	if config.Stop != 0 && config.Interval != 0 && config.Interval < config.Stop {
-		maxIterations *= config.Stop / config.Interval
+		maxIterations *= int(1 + math.Ceil(float64(config.Stop)/float64(config.Interval)))
```
julz: Im a little worried that the new test for this implies that the old test was actually incorrect (because we expect a different output with the same input as before). Could we have a test that explicitly tests the new case (non-dividable values) and make sure we figure out the expected output first, not based on the actual results.

benchmarker/benchmarker_test.go
```diff
@@ -75,7 +75,7 @@ var _ = Describe("Benchmarker", func() {
 		It("repeats a function at n seconds interval and stops at s second", func() {
 			var total int = 0
 			n := 2
-			s := 11
+			s := 10
```
julz: Im not exactly sure what the bug we fixed here was after reading through a couple times; to make it a bit more clear could we add an explicit test for this edge case (i.e. one where the description is something like "it copes with <description of edge case>")

jberkhahn:
Update is up, added additional tests

benchmarker/benchmarker_test.go
```diff
@@ -104,6 +104,14 @@ var _ = Describe("Benchmarker", func() {
 			Execute(RepeatEveryUntil(n, s, func(context.Context) { total += 1 }, nil), workloadCtx)
 			Ω(total).Should(Equal(1))
 		})
+
+		It("runs a function n/s + 1 times if s is a multiple of n", func() {
```
julz: is this a typo? the test says "n/s + 1", the assert says "s/n + 1".. more to the point could we put this in understandable words (e.g. "runs the function as many times as possible until the stop time, given the interval time")

jberkhahn:
Update fixing names and wording in all the repeat every until tests to be more readable. I want to specifically enumerate how many times the experiment is to be repeated, rather than writing something vague like "repeats as much as possible".

julz:
lgtm

PR #99, opened 2014-05-28
Added error reporting for missing & incorrect config files

jberkhahn:
Changed config.go so it propagates errors upward rather than panicking. Added tests to ensure useful errors get reported.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -3,6 +3,7 @@ package config_test
 import (
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	. "github.com/cloudfoundry-incubator/pat/config"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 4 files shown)


config/config_test.go
```diff
@@ -209,4 +210,47 @@ var _ = Describe("ConfigAndFlags", func() {
...
+		It("Returns a useful error when passed an improperly formatted config file", func() {
+			tempfile := filepath.Join(os.TempDir(), "test.yml")
+			configFile, err := os.Create(tempfile)
+			if err != nil {
+				Ω(err).ShouldNot(HaveOccurred())
+			}
+
+			configFile.WriteString("server false\n")
+			configFile.WriteString("iterations: 5\n")
+			configFile.WriteString("concurrency: 1\n")
```
julz, line 241: do we need every single one of these lines?

main.go
```diff
@@ -18,13 +18,17 @@ func main() {
...
+	}
 
 	if useServer == true {
 		logs.NewLogger("main").Info("Starting in server mode")
 		server.Serve()
 	} else {
-		err := cmdline.RunCommandLine()
+		err = cmdline.RunCommandLine()
 		if err != nil {
 			fmt.Println(err)
```
julz, line 33: maybe exit with an error code here too?

jberkhahn:
changes up. added error codes, shorted the cruft in the test, and added a section to the readme detailing the error codes.

PR #108, opened 2014-08-01
remove package.json as it tricks CF into thinking it is a node.js app

drnic:


julz:
I think this breaks travis

drnic:
Sorry; it can probably be removed. I now realise that it's not important to deploy pat CF. It's really a local tool even in -server mode.

PR #123, opened 2015-04-23
README.md - Repeated line removed.

krishnamk00:
This line is repeating and so removed:

 5) See [CF CLI](https://github.com/cloudfoundry/cli) for instructions on installing `cf`
     see [CF Cli](http://github.com/cloudfoundry/cli) for instructions on installing the cloudfoundry cli)."


krishnamk00:
Will review the whole file later instead of single line change.

PR #124, opened 2015-04-23
config-template adding all parameters and their default values.

krishnamk00:
In this config file all the parameters are added with the default values.
All default values are pulled from the code where it is mentioned.
It may be useful for any newbies to get start running PAT quickly with this file.

In the current file, there is a parameter mentioned "-output" and does not seems like existing (let me know if I am wrong)


krishnamk00:
Will submit with the code change for checking...

PR #125, opened 2015-04-27
config file change for default values and not to panic for unknown parameter

krishnamk00:
In this config file all the parameters are added with the default values.
All default values are pulled from the code where it is mentioned.
it is easy to get start running PAT quickly with this file.

In the current file, there is a parameter mentioned "output" and does not seems like existing. 
This creates panic. There is check inserted to ignore unknown parameter and instead take the default.

### DIFF
--- a/config-template.yml
+++ b/config-template.yml
@@ -1,9 +1,28 @@
-# cli_commands are used to specify what the cli commands should be.
-server: false            # true/false should we be running in server mode?
-iterations: 5            # The number of times we wish to itterate over the workload
-concurrency: 1           # The nuber of users we want to simulate running the iterations
-silent: false            # true/false do we want to output command line information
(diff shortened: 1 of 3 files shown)


jberkhahn:
Hey there,
This looks pretty cool, would you mind adding a test for it though? Something in config_test that tests an incorrectly provided param, maybe?

jberkhahn:
Hey @krishna-mk 

Were you interested in finishing these? If we don't hear from you soon we're going to reject these.

krishnamk00:
Hi Jonathan,

Yes, we are very interested in PAT and we are running it in our setup. 
After the cloud foundry summit, this falls apart from the plate. 
I will do it all those pending items in couple of days. Thanks a lot for checking back with me.

Issue #114, opened 2014-08-22
PAT panics if there are any unexpected files in output/csvs/


anEXPer:
```
./pat -config=local-config.yml
panic: runtime error: index out of range

goroutine 1 [running]:
runtime.panic(0x2ef800, 0x6d97b7)
    /usr/local/Cellar/go/1.2.2/libexec/src/pkg/runtime/panic.c:266 +0xb6
github.com/cloudfoundry-incubator/pat/store.(*CsvStore).LoadAll(0xc21000aba0, 0xc210058580, 0x5, 0x8, 0x0, ...)
    /Users/pivotal/go/src/github.com/cloudfoundry-incubator/pat/store/csv.go:190 +0x3d3
github.com/cloudfoundry-incubator/pat/laboratory.(*lab).reload(0xc21006b120)
    /Users/pivotal/go/src/github.com/cloudfoundry-incubator/pat/laboratory/laboratory.go:37 +0x2f
github.com/cloudfoundry-incubator/pat/laboratory.NewLaboratory(0x8137d8, 0xc21000aba0, 0x1, 0x1)
    /Users/pivotal/go/src/github.com/cloudfoundry-incubator/pat/laboratory/laboratory.go:32 +0x8e
github.com/cloudfoundry-incubator/pat/cmdline.func·002(0x8137d8, 0xc21000aba0, 0x0, 0x0)
    /Users/pivotal/go/src/github.com/cloudfoundry-incubator/pat/cmdline/cmdline.go:131 +0x43
github.com/cloudfoundry-incubator/pat/cmdline.func·006(0x8137d8, 0xc21000aba0, 0x8137d8, 0xc21000aba0)
    /Users/pivotal/go/src/github.com/cloudfoundry-incubator/pat/cmdline/cmdline.go:66 +0xcc
github.com/cloudfoundry-incubator/pat/store.WithStore(0xc21000ab60, 0xc21000ab60, 0x98ba0)
    /Users/pivotal/go/src/github.com/cloudfoundry-incubator/pat/store/config.go:32 +0xd6
github.com/cloudfoundry-incubator/pat/cmdline.func·007(0xc210000398, 0xc210044a50)
    /Users/pivotal/go/src/github.com/cloudfoundry-incubator/pat/cmdline/cmdline.go:82 +0x60
github.com/cloudfoundry-incubator/pat/cmdline.validateParameters(0x813748, 0xc210000398, 0x853dc0, 0x8137b0, 0xc210000398)
[...]

PR #126, opened 2015-04-27
Fix issue: PAT panics if there are any unexpected files in output/csvs/ #114

krishnamk00:
PAT panics if there are any unexpected files in output/csvs. This is because PAT is looking for standard file format it those items. Added a fix to ignore those elements and avoid the panic. There can be custom file created there from "logging:file" parameter passed and so that check is not always applicable to all files.

### DIFF
--- a/store/csv.go
+++ b/store/csv.go
@@ -189,13 +189,17 @@ func (store *CsvStore) LoadAll() (samples []experiment.Experiment, err error) {
 	samples = make([]experiment.Experiment, 0)
 	for _, f := range files {
 		base := strings.Split(f.Name(), ".")[0]
-		name := strings.SplitN(base, "-", 2)[1]
-		if len(name) > 0 {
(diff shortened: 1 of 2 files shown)


jberkhahn:
Hey, this looks like a neat idea, but the way you implemented it seems kind of brittle - it just checks all the files that happen to have a hyphen in the name. Wouldn't it be a better idea to try and grab all files that match the _-_-_-_-*.csv pattern that PAT outputs? You also didn't include a test for the feature.

krishnamk00:
Hi Jo, I added .csv file check and the test code to validate . So far looks good in my run, If any output file with no .csv and or any other format, those files will be ignored. The PAT getting panic if any arbitrary file in the output/csvs folder is no more an issue.  Let me know if you think any more additions to this fix. I will work on other fixes too. - thanks : Krishna

PR #127, opened 2015-04-27
WokLoad List desc modified to list workload target display

krishnamk00:
Currently if you run 'pat -list-workloads' it shows all the descriptions of the workloads but does not really show what work load to be passed. After this modifications the actual workload list target along with description will be shown.


jberkhahn:
Hey there, I don't really think this is necessary, as the workload titles themselves are the actual workload arguments that you enter. It seems just adding them again is just needless repetition.

I'm going to close this one, but feel free to comment if you think this is really needed.

Issue #115, opened 2014-08-22
Setting a custom file name for output does nothing


anEXPer:
I get an output file that's named like `1408741776408079642-23429aeb-fd80-4823-755d-cc47e88dd963.csv` regardless of any efforts to attain a different outcome. 

I can specify `-output="~/workspace/test.csv"` or just `-output="test.csv"`, same result. I can put it in the config file as `output: "test.csv"` (or the absolute path) and it does the same thing.

This not working is a significant reason I've found it necessary to wrap PAT in a script; I need to be able to organize and name output effectively.

potatoronin:
https://www.pivotaltracker.com/story/show/77479712
"Sorry, we couldn't find that story"

PR #128, opened 2015-04-27
Fix issue: Setting a custom file name for output does nothing #115

krishnamk00:
logging:file parameter passed was not taking as output file instead always PAT takes the default output file format always.  This fix will:
- Take the argument passed a output file and place it in the correct output dir.
- If file is passed as dir/file format it checks the format and process the file, else it used to create panic.

### DIFF
--- a/logs/config.go
+++ b/logs/config.go
@@ -3,6 +3,7 @@ package logs
 import (
 	"os"
 	"strings"
+	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


jberkhahn:
Hey there,

This looks like a good idea, but it's currently broken - you missed an ending } in logs/config.go, so it won't even compile. You also didn't write any tests describing the new feature - think you could add some?

